PR #6, opened 2021-11-19
Implement Speed Modifier for Farmland

mixxit:
This is based on the sapling modifier


mixxit:
Tyron
i had a quick look, but i'm a bit uncertain if that is the right solution, regarding the configurable growth times
the vs team has long been talking about making crop growth season based, i.e. you can only grow 1-2 time a year, but up the drop rate accordingly

Tyron — Today at 11:57 AM
you can also patch those growthimes during runtime in code, too
if you want a single value to tweak

Closing this in favour of a mod approach

VinterNacht:
It seems like it would already be seasonal growth if you extended the growth times.  I can say that given the current temperature sensitivities/requirements of the crops they already work in the manner you describe on an extended growth period.  They become seasonal because only certain seasons in certain climates at certain altitudes provide the required conditions.

PR #9, opened 2022-03-28
allow poultice to heal others

mixxit:
is it possible we can have something like this to heal others with bandages?


tyronx:
definitely can be a thing but should come with its own first person animation as well as an entity interaction help (which doesn't exist yet)

Craluminum2413:
How tooltip for entity harvesting thing works then?

pizza2004:
This PR is very old and the code has changed drastically.  Feel free to submit a new one with these changes applied against the newer code if desired.

PR #10, opened 2022-03-31
Make BlockEntityOven public for ease in modding

00500005:
Most block entities are public so I just assume that this was a slight oversight.

Also, seems github was insistent that line endings be updated. Feel free to cancel this change and make a commit yourself if it bothers you.


tyronx:
ah yes, thats not on purpose. I've fixed that now. thanks for the suggestion

PR #15, opened 2022-11-21
Set `/wgen del` to delete around player, to match Wiki, `/wgen regen`, and `/wgen delrock`

Kevok93:
https://wiki.vintagestory.at/index.php/List_of_server_commands#/wgen

/wgen del [radius]
Remove chunks around the player


tyronx:
thanks for the fix, but I've already added this to 1.18

PR #16, opened 2022-12-06
Fix anegostudios/VintageStory-Issues#1279

noelle-lavenza:
Fixes https://github.com/anegostudios/VintageStory-Issues/issues/1279.
Replaces voxelPos in the bounds check with offPos.

This doesn't fix the issue of the radius being set by cave tools, but should fix the out of bounds crash.


tyronx:
oh, thanks! I'll include the fix manually. 👍

PR #17, opened 2022-12-07
Fix behavior methods never get called

Craluminum2413:
Related
https://github.com/anegostudios/VintageStory-Issues/issues/2305

### DIFF
--- a/BlockEntity/BECrate.cs
+++ b/BlockEntity/BECrate.cs
@@ -67,7 +67,7 @@ namespace Vintagestory.GameContent
 
                 if (inventory.Empty) return preferredLidState;
                 var stack = inventory.FirstNonEmptySlot.Itemstack;
-                if (stack == null || (stack.ItemAttributes != null && stack.ItemAttributes["inContainerTexture"].Exists)) return preferredLidState;
+                if (stack?.Collectible == null || (stack.ItemAttributes != null && stack.ItemAttributes["inContainerTexture"].Exists)) return preferredLidState;
 
                 bool? displayInsideCrate = stack.ItemAttributes?["displayInsideCrate"].Exists != true ? null : stack.ItemAttributes?["displayInsideCrate"].AsBool(true);
                 bool hasContentTexture = (stack.Block != null && stack.Block.DrawType == EnumDrawType.Cube && displayInsideCrate != false) || displayInsideCrate == true;
--- a/BlockEntity/BESpawner.cs
+++ b/BlockEntity/BESpawner.cs
@@ -400,6 +400,23 @@ namespace Vintagestory.GameContent
         public override void OnBlockRemoved()
         {
             base.OnBlockRemoved();
+
+            var sapi = Api as ICoreServerAPI;
+            if (sapi != null)
+            {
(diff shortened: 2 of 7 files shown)


Block/BlockWateringCan.cs
```diff
@@ -88,6 +93,7 @@ public override void OnHeldInteractStart(ItemSlot slot, EntityAgent byEntity, Bl
             }
 
             handHandling = EnumHandHandling.PreventDefault;
+            base.OnHeldInteractStart(slot, byEntity, blockSel, entitySel, firstEvent, ref handHandling);
```
tyronx, line 96: this one looks out of place

Block/BlockWateringCan.cs
```diff
@@ -258,7 +264,7 @@ public override bool OnHeldInteractStep(float secondsUsed, ItemSlot slot, Entity
                 byEntity.World.SpawnParticles(WaterParticles, byPlayer);
             }
 
-            return true;
+            return base.OnHeldInteractStep(secondsUsed, slot, byEntity, blockSel, entitySel);
```
tyronx, line 267: this one also looks like it would break things

BlockEntity/BEPitKiln.cs
```diff
@@ -499,6 +499,8 @@ public override void GetBlockInfo(IPlayer forPlayer, StringBuilder dsc)
 
             if (Lit) dsc.AppendLine(Lang.Get("Lit"));
             else dsc.AppendLine(Lang.Get("Unlit"));
+
+            base.GetBlockInfo(forPlayer, dsc);
```
tyronx, line 503: same here, this would display additional information

Item/ItemStone.cs
```diff
@@ -69,18 +71,26 @@ public override void OnHeldInteractStart(ItemSlot itemslot, EntityAgent byEntity
                     {
                         (this.api as ICoreClientAPI).TriggerIngameError(this, "toosoft", Lang.Get("This type of stone is too soft to be used for knapping."));
                     }
+
+                    base.OnHeldInteractStart(slot, byEntity, blockSel, entitySel, firstEvent, ref handling);
```
tyronx, line 75: same issue here

tyronx:
the other parts are now integrated

BlockEntity/BEPitKiln.cs
Craluminum2413: BlockEntityBehavior can't override GetBlockInfo for pit kilns due to this

Block/BlockWateringCan.cs
Craluminum2413: BlockBehavior doesn't call OnHeldInteractStep because of this, which makes custom interactions impossible

Block/BlockWateringCan.cs
Craluminum2413: lockBehavior doesn't call OnHeldInteractStart because of this, which also makes custom interactions impossible

tyronx: you'd have multiple interactions running at the same time with this enabled, its not meant for this. It would create undefind ingame behavior. If you want to replace a blocks interact behavior you'll have to patch it

PR #19, opened 2023-06-01
Explicitly stated ":G" formatting for liquid content, which should fix the current display bug

Moby4585:
That fixed the liquid content display for Lavoisier, I guess it should work for Vanilla too
(fixes the bug where it displays ",1 litres of water" when you have 0.16L for instance, ignoring the zero and 0.01L portions, and just displays " litres of water" with no quantity when you have less than 0.05L)

EDIT: It should probably also be in GetPlacedBlockInfo for an actual fix :D

### DIFF
--- a/Systems/Liquid/BlockLiquidContainerBase.cs
+++ b/Systems/Liquid/BlockLiquidContainerBase.cs
@@ -225,10 +225,10 @@ namespace Vintagestory.GameContent
             
             if (litres == 1)
             {
-                return Lang.Get("{0} ({1} litre of {2})", inSlot.Itemstack.GetName(), litres, incontainername);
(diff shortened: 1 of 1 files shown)


Th3Dilli:
hey thank you very much for your PR

we had a look at it and had to change our internal handling of plural support for example in this case for the liquids
(1 litre, 2 litres 2.23 litres etc)

I can not promise when this will hit the main game but at least when 1.19 will be released maybe with the next release of 1.18 already


to clarify we will not merge any PR's here since this is an external public clone of our internal repo
but we highly appreciate it when we get PR's since that cuts down the time to fix things if it is ready as is we will merge it internally and give credit in the patch notes for it and eventually on the game credits to where more enormous contributions where made

PR #21, opened 2023-07-25
Fix loading of rift ward in the off state

korobya:
This PR fixes an issue with loading of rift wards in off state when loading the world.

Discord report: https://discord.com/channels/302152934249070593/1132917500720467969


Th3Dilli:
thank you for the report

it will be fixed with an upcoming release

(note: we use an internal report and push the changes here later when released that's why things don get merged but we highly appreciate any PR's )

PR #22, opened 2023-07-29
Fix glider behaviour in backpack's content

korobya:
This PR fixes the issue that caused the glider to also work while in the contents of the backpack.

There was also a typo in the name of the property. I fixed it too.

As a suggestion - maybe make `ItemSlot[] backPackSlots` and `List<ItemSlot> backPackContents` in `InventoryPlayerBackPacks.cs` public to separate the iteration by "backpack-only" slots and by the contents of backpacks?

Discord report: https://discord.com/channels/302152934249070593/1134876018763649164


tyronx:
thanks! will be in a future release

PR #23, opened 2023-07-29
Make it so that terminus teleporter has no fuel by default

korobya:
This PR makes it so that terminus teleporter has no fuel by default. So that it cannot be "recharged" by breaking it and placing it again.

Discord report: https://discord.com/channels/302152934249070593/1134930277173567499


tyronx:
thanks a lot! Will be in the next release.

PR #24, opened 2023-08-25
Fix a bug causing lag and NREs on the server

Dmitry221060:
Fixes anegostudios/VintageStory-Issues#3041 by adding a null check


tyronx:
thanks a lot, will manually merge this and include it in the next release!

PR #25, opened 2023-10-31
Fix signs having the same texture as first placed sign

Craluminum2413:


Craluminum2413:
First placed sign was modded **acacia sign**
image

tyronx:
thanks, will be in pre5

PR #26, opened 2023-10-31
Move carvingTime boolean from method to field

Craluminum2413:


tyronx:
👍

Craluminum2413:
Probably it was bad idea to make it static readonly, seems like I can not longer patch carvingTime by using Reflection or Harmony, can you make it getter?

tyronx:
yea ok

tyronx:
i'll just remove the `readonly` part

PR #28, opened 2023-12-08
Seaweedgoesinsea

spooq:
The LiquidCode checks for seaweed worldgen are incorrect. Also the seaweed is far too short when it's actually placed in the ocean. I've based the lengths and depth on giant kelp and bull kelp stats.

To have seaweed actually go in the ocean:

  {
    "file": "game:worldgen/blockpatches/other.json",
    "op": "add",
    "path": "/4/placement",
    "side": "Server",
    "value": "UnderSeaWater"
  },
  {
    "file": "game:worldgen/blockpatches/other.json",
    "op": "add",
    "path": "/5/placement",
    "side": "Server",
    "value": "UnderSeaWater"
  }

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
 /.gitignore
 /obj
+/.vs
+/.gitignore
+/VSSurvivalMod.csproj
+/VSSurvivalMod.sln
+/.gitignore
+/VSSurvivalMod.csproj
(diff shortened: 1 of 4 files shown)


pizza2004:
From what I can tell this has been fixed another way in the interim and this PR is no longer necessary.  If there is still an issue feel free to submit a new PR.

PR #30, opened 2023-12-14
Fix squeezing honeycomb into ground storage slot

Craluminum2413:


tyronx:
thanks, this has been fixed for rc6

PR #31, opened 2024-01-05
Fix character class gear issue

Craluminum2413:
https://github.com/anegostudios/VintageStory-Issues/issues/3390


tyronx:
thanks, will be in rc6

PR #32, opened 2024-01-14
Allow to place rope since it got GroundStorable behavior

Craluminum2413:


tyronx:
thank, will include this

PR #35, opened 2024-02-01
Fix peat brick interaction for behaviors

Craluminum2413:


Craluminum2413:
@BillyGalbreath for example, when you want to add interaction with entities, like "right-click" on entity and put fuel into it

Craluminum2413:
Peat piles are ground storable now

PR #36, opened 2024-02-10
Add support for multiple parts itemcode for ItemPlantableSeed

Craluminum2413:
Currently it supports only seeds itemtype with `xxx-yyy` itemcode, with this change it will support seeds itemtype with any amount of parts as well, for example `xxx-yyy-yyy` or `xxx-yyy-yyy-yyy`. 

Shouldn't break anything since last code part only matters in `crop-yyy-stage` blockcode to check for current stage.


pizza2004:
This has been handled by updating them to use `Variant` instead.

PR #37, opened 2024-02-10
Call base method in OnHeldInteractStart in ItemPlantableSeed.cs

Craluminum2413:
Allows mods to add their own interactions to ItemPlantableSeed through CollectibleBehavior


pizza2004:
Will be in future release.

PR #39, opened 2024-02-20
Add missing tooltips for ItemFirestarter

Craluminum2413:


pizza2004:
Will be in future release.

PR #41, opened 2024-02-26
Don't hardcode ItemStone damage

Craluminum2413:


pizza2004:
Thanks, this is now merged.

PR #43, opened 2024-02-29
Add wildcard and regex support for BehaviorUnstableFalling exceptions

Craluminum2413:
Fixes https://github.com/anegostudios/VintageStory-Issues/issues/1830


pizza2004:
Will be in a future version.

PR #46, opened 2024-03-07
(Ground Storage) Apply smoke particles to some items

Craluminum2413:
This change fixes https://github.com/anegostudios/VintageStory-Issues/issues/1577

Crucible, cooking pot and meal containers don't have particles when placed, this code adds those particles.

I tested all particle positions, they are aligned with MeshAngle.

I couldn't use tfMatrices because it is a mess when rotated, -8.xxxx, -4.xxx etc. are unreadable values so it is much easier to do it manually.

Top position is set from MaxY field from first collision box, so you might need to lower collision box in GroundStorable behavior for some items or add OverrideMaxYParticlePos to GroundStorageProperties and then apply it in OnAsyncClientParticleTick method.

### DIFF
--- a/Block/BlockSmeltedContainer.cs
+++ b/Block/BlockSmeltedContainer.cs
@@ -485,7 +485,7 @@ namespace Vintagestory.GameContent
             stack.Attributes.SetInt("units", units);
         }
 
-        KeyValuePair<ItemStack, int> GetContents(IWorldAccessor world, ItemStack stack)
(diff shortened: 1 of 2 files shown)


pizza2004:
Could you fix this so there aren't merge conflicts?

Craluminum2413:
No idea how to fix that

pizza2004:
Thank you, this was a useful basis to reimplement this feature.  Will be in a future version.

PR #49, opened 2024-03-09
Make wooden pan GroundStorable

Craluminum2413:
Replace `survival/blocktypes/wood/pan.json` behaviors:
```json
{ "name": "GroundStorable", "properties": { "layout": "SingleCenter" } },
{ "name": "Unplaceable", "__comment": "The ground storable obsoletes this being a block. Should be an item, but is kept a block for backwards compatibility" },
{ "name": "RightClickPickup" }
```

Pan content is not shown when in ground storage because BlockPan class doesn't have IContainedMeshSource interface implemented (I may later include it to pull request)

image

Fixes two issues:
https://github.com/anegostudios/VintageStory-Issues/issues/1332
https://github.com/anegostudios/VintageStory-Issues/issues/1174


pizza2004:
Could you possibly implement the IContainedMeshSource part now?

pizza2004:
> @pizza2004 what for?

I don't think this should be a vanilla feature if the contents are invisible is why. :P

Craluminum2413:
@pizza2004 Contents are invisible either way

pizza2004:
> @pizza2004 Contents are invisible either way, just place it and you will see

Are you willing to try and make them not invisible, then?

Craluminum2413:
> > @pizza2004 Contents are invisible either way, just place it and you will see
> 
> Are you willing to try and make them not invisible, then?

Yes

Craluminum2413:
Damn i broke commits completely

pizza2004:
Will be in pre.2!  Thanks!

PR #50, opened 2024-03-10
(Ground Storage) Easier interactions

Craluminum2413:
When interacting with ground storage with empty hand, select nearby non-empty slot if selected slot is empty.
It may reduce a lot of confusion when player tries to pick up something but it doesn't get picked up because player is aiming at empty slot.

Mod compiled for 1.19.5-rc.1: [TestMod.zip](https://github.com/anegostudios/vssurvivalmod/files/14550773/TestMod.zip)

Concept image:
image

Video demonstration:
https://github.com/anegostudios/vssurvivalmod/assets/69315569/936f8e6f-d8e7-450d-ba28-74cb4a929d31

### DIFF
--- a/BlockEntity/BEGroundStorage.cs
+++ b/BlockEntity/BEGroundStorage.cs
@@ -468,19 +468,15 @@ namespace Vintagestory.GameContent
 
                     case EnumGroundStorageLayout.WallHalves:
                     case EnumGroundStorageLayout.Halves:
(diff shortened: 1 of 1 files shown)


pizza2004:
It seems to me like this would create equally as much or more confusion as the current behavior and I can also think of what I believe would be better solutions to the confusion, but thank you for the submission.

PR #57, opened 2024-04-21
Fix auction total cost description after changing delivery mode

quppeq:
Updated description after changing delivery mode


tyronx:
oh, right, i noticed this one too. this is now fixed 1.20! thanks.

PR #58, opened 2024-04-28
Fix support beam functionality

dmchurch:
Adding or removing support beams doesn't have effect when you load a game that already has support beams in a given chunk. Whoops!

This doesn't address any problems that players will have on loading worlds where the physical support beams don't match the SBD structure, as will happen after being affected by this bug, so this may want to also do some amount of rechecking the physical blocks.

Consider this my job interview, then contract me to fix the support beam code, imo 😄


tyronx:
wow epic fail. thanks a lot for the PR. Will be in 1.20!

PR #59, opened 2024-06-04
Add ability for cooking vessels "cookingContainerSlots" attribute to specify number of cooking container slots

Pursec:
Current vanilla implementation ignores the value set by the objects attribute and always creates a grid of 4 slots regardless of the attributes value.

PR is for a basic implementation/adjustment to expand the num of cooking slots in a smeltingInventory to 8 maximum, with the "cookingContainerSlots" attribute adjusting the range of available slots returned by the CookingSlots getter.

A minor adjustment to the Firepit Gui was added to support quantity of slots non-divisible cleanly by 4 to still adjust the bounds of the cookingSlotsSlotBounds correctly for the amount of rows needed. This does not adjust the firepit heating graphic currently which gets cut off by the expanded bounds when the cookingSlot rows exceed 1.

### DIFF
--- a/Gui/GuiDialogBlockEntityFirepit.cs
+++ b/Gui/GuiDialogBlockEntityFirepit.cs
@@ -4,6 +4,7 @@ using Vintagestory.API.Datastructures;
 using Vintagestory.API.MathTools;
 using Vintagestory.API.Common;
 using Vintagestory.API.Config;
+using System;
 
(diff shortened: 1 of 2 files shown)


pizza2004:
Since this is in conflict and has some bugs I'm going to close it, but if you wish to try and fix it we can discuss how and it can be reopened.

PR #60, opened 2024-08-20
Fix Berry Bush Animal Harvesting and Pruning Reset Bugs

pizza2004:
Fixed a anegostudios/VintageStory-Issues#4034 and anegostudios/VintageStory-Issues#3811 as well as adding a proper interaction tip for pruning.  A new `exchangeBlock: true` attribute entry needs to be added to Harvest behavior in the smallberrybush and largeberrybush JSON files to go along with this, to make sure it works fully.

### DIFF
--- a/BlockBehavior/BehaviorHarvestable.cs
+++ b/BlockBehavior/BehaviorHarvestable.cs
@@ -14,7 +14,6 @@ namespace Vintagestory.GameContent
 
         AssetLocation harvestedBlockCode;
         Block harvestedBlock;
-        bool exchangeBlock;
         string interactionHelpCode;
 
         public BlockBehaviorHarvestable(Block block) : base(block)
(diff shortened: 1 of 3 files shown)


BlockBehavior/BehaviorHarvestable.cs
```diff
@@ -122,7 +122,8 @@ public override void OnBlockInteractStop(float secondsUsed, IWorldAccessor world
 
                 if (harvestedBlock != null)
                 {
-                    world.BlockAccessor.SetBlock(harvestedBlock.BlockId, blockSel.Position);
+                    if (!properties["exchangeBlock"].AsBool(false)) world.BlockAccessor.SetBlock(harvestedBlock.BlockId, blockSel.Position);
+                    else world.BlockAccessor.ExchangeBlock(harvestedBlock.BlockId, blockSel.Position);
```
pizza2004, line 126: This makes it so you can exchange blocks when harvested so we don’t have to reset the berry bush entities when the block is harvested.

BlockEntity/BEBerryBush.cs
```diff
@@ -188,6 +188,7 @@ private void CheckGrow(float dt)
         public override void OnExchanged(Block block)
         {
             base.OnExchanged(block);
+            transitionHoursLeft = GetHoursForNextStage();
```
pizza2004, line 191: Moving it here makes sure that it always runs when the block gets exchanged, which helps with making sure it runs after harvest now, but also fixes it so raccoons harvesting will properly reset the timer like you would expect.

tyronx:
thanks, this is now merged!

PR #61, opened 2024-08-25
Overhaul Fruitpress to Fix Bugs and Improve Use

pizza2004:
This is a slight rewrite of how the Fruitpress functions, fixing a few small bugs like anegostudios/VintageStory-Issues#3868 (producing dry mash that won’t properly stack), or allowing you to add dry mash that you would then lose forever.  It also makes it a little easier to use and feel a little more responsive, like making it so you can stop squeezing the fruit halfway through and it will actually stop squeezing at that amount.  You could probably increase the speed of squeezing much higher than 50f on line 238 and have it feel like the juice squeezed out in time with your cranking, so letting go of the mouse button would immediately stop, as well, if that were desired.

I’ve added copious comments so it’s obvious what all the added lines are meant to do, in case there’s a desire to change some part of it, as well.  It also appears that whatever issue was happening where the server would play the animation at half the speed is no longer an issue so I removed all of the code related to that and it looks much smoother now.

The main thing I couldn’t crack was how to make the crank and the mash always look like they are touching.  The selfHeight portion of the code does a good job but it’s not quite exactly right and in all my tinkering I couldn’t find a simple way to make it look more consistent without needing to program specific cases for each mash height.

### DIFF
--- a/BlockEntity/BEFruitpress.cs
+++ b/BlockEntity/BEFruitpress.cs
@@ -1,5 +1,6 @@
 ﻿using ProtoBuf;
 using System;
+using System.Collections.Generic;
 using System.Text;
 using Vintagestory.API.Client;
 using Vintagestory.API.Common;
@@ -26,11 +27,19 @@ namespace Vintagestory.GameContent
         Screw
     }
 
+    public enum EnumFruitPressAnimState
+    {
+        ScrewStart,
+        Unscrew,
+        ScrewContinue
+    }
+
     [ProtoContract(ImplicitFields = ImplicitFields.AllPublic)]
     public class FruitPressAnimPacket
     {
-        public bool AnimationActive;
+        public EnumFruitPressAnimState AnimationState;
         public float AnimationSpeed;
+        public float CurrentFrame = 0;
     }
 
 
@@ -90,10 +99,17 @@ namespace Vintagestory.GameContent
             EaseInSpeed = 3
         };
 
+        float? loadedFrame;
+        bool serverListenerActive;
 
         long listenerId;
         double juiceableLitresCapacity = 10;
+        double screwPercent;
+        double squeezedLitresLeft;
+        double pressSqueezeRel;
         bool squeezeSoundPlayed;
+        bool squeezedJuiceLeft;
+        int dryStackSize = 0;
 
         public ItemSlot MashSlot => inv[0];
         public ItemSlot BucketSlot => inv[1];
@@ -115,7 +131,7 @@ namespace Vintagestory.GameContent
         {
             get
             {
-                return mashStack.Attributes.GetDouble("juiceableLitresTransfered");
+                return mashStack?.Attributes?.GetDouble("juiceableLitresTransfered") ?? 0;
             }
             set
             {
@@ -134,8 +150,8 @@ namespace Vintagestory.GameContent
         public bool CompressAnimActive => animUtil.activeAnimationsByAnimCode.ContainsKey("compress") || animUtil.animator.GetAnimationState("compress")?.Active == true;
 
 
-        public bool CanScrew => !CompressAnimActive || compressAnimMeta.AnimationSpeed == 0;
-        public bool CanUnscrew => CompressAnimFinished;
+        public bool CanScrew => !CompressAnimFinished;
+        public bool CanUnscrew => CompressAnimFinished || CompressAnimActive;
         public bool CanFillRemoveItems => !CompressAnimActive;
 
         public BlockEntityFruitPress()
@@ -171,6 +187,11 @@ namespace Vintagestory.GameContent
                     renderer.reloadMeshes(getJuiceableProps(mashStack), true);
                     genBucketMesh();
                 }
+                else if (serverListenerActive)
+                {
+                    animUtil.StartAnimation(compressAnimMeta);
+                    if (listenerId == 0) listenerId = RegisterGameTickListener(onTick100msServer, 25);
+                }
             }
         }
 
@@ -180,8 +201,9 @@ namespace Vintagestory.GameContent
         private void onTick25msClient(float dt)
         {
             double squeezeRel = mashStack?.Attributes.GetDouble("squeezeRel", 1) ?? 1;
+            float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
 
-            if (MashSlot.Empty || renderer.juiceTexPos == null || squeezeRel >= 1) return;
+            if (MashSlot.Empty || renderer.juiceTexPos == null || squeezeRel >= 1 || pressSqueezeRel > squeezeRel || squeezedLitresLeft < 0.01) return;
 
             var rand = Api.World.Rand;
 
@@ -198,7 +220,7 @@ namespace Vintagestory.GameContent
 
                 minPos.Mul(8 / 16f);
                 maxPos.Mul(8 / 16f);
-                maxPos.Y = 5 / 16f - (1 - squeezeRel) * 0.25f;
+                maxPos.Y = 5 / 16f - (1 - squeezeRel + Math.Max(0, 0.9f - selfHeight)) * 0.5f;
 
                 minPos.Add(face.Normalf.X * 1.2f / 16f, 0, face.Normalf.Z * 1.2f / 16f);
                 maxPos.Add(face.Normalf.X * 1.2f / 16f, 0, face.Normalf.Z * 1.2f / 16f);
@@ -224,23 +246,39 @@ namespace Vintagestory.GameContent
 
         private void onTick100msServer(float dt)
         {
+            RunningAnimation anim = animUtil.animator.GetAnimationState("compress");
+            if (serverListenerActive)
+            {
+                anim.CurrentFrame = loadedFrame ?? 0;
+                updateSqueezeRel(animUtil.animator.GetAnimationState("compress"));
+                serverListenerActive = false;
+                loadedFrame = null;
+                return;
+            }
+            else if (CompressAnimActive) (Api as ICoreServerAPI)?.Network.BroadcastBlockEntityPacket(Pos, PacketIdAnimUpdate, new FruitPressAnimPacket() { AnimationState = EnumFruitPressAnimState.ScrewContinue, AnimationSpeed = compressAnimMeta.AnimationSpeed, CurrentFrame = anim.CurrentFrame });
             if (MashSlot.Empty) return;
 
             var juiceProps = getJuiceableProps(mashStack);
             double totalHours = Api.World.Calendar.TotalHours;
 
             double squeezeRel = mashStack.Attributes.GetDouble("squeezeRel", 1);
-            double litresToTransfer = Math.Min(juiceableLitresLeft, (totalHours - lastLiquidTransferTotalHours) * 50f);
 
-            if (Api.Side == EnumAppSide.Server && squeezeRel < 1 && !squeezeSoundPlayed && juiceableLitresLeft > 0)
+            // First we need to calculate how squeezed down the mash has been and therefore how much we're allowed to take
+            squeezedLitresLeft = Math.Max(Math.Max(0, squeezedLitresLeft), juiceableLitresLeft - ((juiceableLitresLeft + juiceableLitresTransfered) * screwPercent));
+            double litresToTransfer = Math.Min(squeezedLitresLeft, Math.Round((totalHours - lastLiquidTransferTotalHours) * (CompressAnimActive ? GameMath.Clamp(squeezedLitresLeft * (1 - squeezeRel) * 500f, 25f, 100f) : 5f), 2));
+
+            if (Api.Side == EnumAppSide.Server && CompressAnimActive && squeezeRel < 1 && pressSqueezeRel <= squeezeRel && !squeezeSoundPlayed && juiceableLitresLeft > 0)
             {
                 Api.World.PlaySoundAt(new AssetLocation("sounds/player/wetclothsqueeze.ogg"), Pos, 0, null, false);
                 squeezeSoundPlayed = true;
+                squeezedJuiceLeft = true;
             }
 
             BlockLiquidContainerBase cntBlock = BucketSlot?.Itemstack?.Collectible as BlockLiquidContainerBase;
 
-            if (Api.Side == EnumAppSide.Server && squeezeRel < 1 && totalHours - lastLiquidTransferTotalHours > 0.01)
+            // Since the sound only plays once when the screw touches the mash we can reuse the variable here to make
+            // sure this runs as long as there is still juice left to drip out, even if the player releases the screw
+            if (Api.Side == EnumAppSide.Server && squeezedLitresLeft > 0 && squeezedJuiceLeft)
             {
                 ItemStack liquidStack = juiceProps.LiquidStack.ResolvedItemstack;
                 liquidStack.StackSize = 999999;
@@ -263,28 +301,16 @@ namespace Vintagestory.GameContent
                 }
 
                 juiceableLitresLeft -= actuallyTransfered;
+                squeezedLitresLeft -= pressSqueezeRel <= squeezeRel ? actuallyTransfered : (actuallyTransfered * 20); // Let the mash drain less if screw is released.
                 juiceableLitresTransfered += actuallyTransfered;
                 lastLiquidTransferTotalHours = totalHours;
                 MarkDirty(true);
             }
-
-
-            if (juiceableLitresLeft <= 0.01)
+            else if (Api.Side == EnumAppSide.Server && (!CompressAnimActive || juiceableLitresLeft <= 0))
             {
-                // Hack to fix rounding errors
-                if (cntBlock != null)
-                {
-                    float litres = cntBlock.GetCurrentLitres(BucketSlot.Itemstack);
-                    cntBlock.SetCurrentLitres(BucketSlot.Itemstack, (float)Math.Round(100 * litres) / 100f);
-                }
-
+                squeezedJuiceLeft = false;
                 UnregisterGameTickListener(listenerId);
                 listenerId = 0;
-
-                int stacksize = GameMath.RoundRandom(Api.World.Rand, (float)juiceableLitresTransfered);
-                mashStack.Attributes.RemoveAttribute("juiceableLitresTransfered");
-                mashStack.Attributes.RemoveAttribute("juiceableLitresLeft");
-                mashStack.StackSize = (int)(stacksize * juiceProps.PressedDryRatio);
 
                 MarkDirty(true);
             }
@@ -301,11 +327,11 @@ namespace Vintagestory.GameContent
         {
             firstEvent |= Api.Side == EnumAppSide.Server;
 
-            if (section == EnumFruitPressSection.MashContainer)
+            if (section == EnumFruitPressSection.MashContainer && firstEvent)
             {
                 return InteractMashContainer(byPlayer, blockSel);
             }
-            if (section == EnumFruitPressSection.Ground)
+            if (section == EnumFruitPressSection.Ground && firstEvent)
             {
                 return InteractGround(byPlayer, blockSel);
             }
@@ -322,36 +348,26 @@ namespace Vintagestory.GameContent
             if (Api.Side == EnumAppSide.Server) return true; // We let the client control this
 
             // Start
-            if (!CompressAnimActive && firstEvent)
+            if (!CompressAnimActive && byPlayer.Entity.Controls.CtrlKey && firstEvent)
             {
-                compressAnimMeta.AnimationSpeed = 0.5f;
-                animUtil.StartAnimation(compressAnimMeta);
-                squeezeSoundPlayed = false;
-
                 (Api as ICoreClientAPI).Network.SendBlockEntityPacket(Pos, PacketIdScrewStart);
-
-                if (listenerId == 0)
-                {
-                    listenerId = RegisterGameTickListener(onTick25msClient, 25);
-                }
 
                 return true;
             }
 
             // Unscrew
-            if (CanUnscrew && firstEvent)
+            if (CanUnscrew && !byPlayer.Entity.Controls.CtrlKey && firstEvent)
             {
-                compressAnimMeta.AnimationSpeed = 1.5f;
-                animUtil.StopAnimation("compress");
                 (Api as ICoreClientAPI).Network.SendBlockEntityPacket(Pos, PacketIdUnscrew);
+
                 return true;
             }
 
             // Continue
-            if (compressAnimMeta.AnimationSpeed == 0)
+            if (compressAnimMeta.AnimationSpeed == 0 && byPlayer.Entity.Controls.CtrlKey)
             {
-                compressAnimMeta.AnimationSpeed = 0.5f;
                 (Api as ICoreClientAPI).Network.SendBlockEntityPacket(Pos, PacketIdScrewContinue);
+
                 return true;
             }
 
@@ -376,9 +392,27 @@ namespace Vintagestory.GameContent
                 var hprops = getJuiceableProps(handStack);
                 if (hprops == null) return false;
 
-                var pressedStack = hprops.PressedStack.ResolvedItemstack.Clone();
-                if (MashSlot.Empty) MashSlot.Itemstack = pressedStack;
-                else if (mashStack.StackSize >= 10)
+                // Don't try to put dry mash back into the machine
+                if (hprops.LitresPerItem == null && !handStack.Attributes.HasAttribute("juiceableLitresLeft")) return false;
+
+                var pressedStack = hprops.LitresPerItem != null ? hprops.PressedStack.ResolvedItemstack.Clone() : handStack.GetEmptyClone();
+                if (MashSlot.Empty)
+                {
+                    MashSlot.Itemstack = pressedStack;
+
+                    // Directly transfer the fruit mash to the machine if it's empty, but only one if it's a stack
+                    if (hprops.LitresPerItem == null)
+                    {
+                        mashStack.StackSize = 1;
+                        dryStackSize = (int)(GameMath.RoundRandom(Api.World.Rand, (float)juiceableLitresLeft + (float)juiceableLitresTransfered) * getJuiceableProps(mashStack).PressedDryRatio);
+                        handslot.TakeOut(1);
+                        MarkDirty(true);
+                        renderer?.reloadMeshes(hprops, true);
+                        (byPlayer as IClientPlayer)?.TriggerFpAnimation(EnumHandInteract.HeldItemInteract);
+                        return true;
+                    }
+                }
+                else if (juiceableLitresLeft + juiceableLitresTransfered >= juiceableLitresCapacity)
                 {
                     (Api as ICoreClientAPI)?.TriggerIngameError(this, "fullcontainer", Lang.Get("Container is full, press out juice and remove the mash before adding more"));
                     return false;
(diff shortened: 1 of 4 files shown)


BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
                 } else
-                {  
-                    float desiredTransferSizeLitres = byPlayer.Entity.Controls.ShiftKey ? (float)hprops.LitresPerItem : Math.Min(handStack.StackSize, 4) * (float)hprops.LitresPerItem;
-                    transferableLitres = (float)Math.Min(desiredTransferSizeLitres, juiceableLitresCapacity - juiceableLitresLeft);
+                {
+                    int desiredTransferAmount = Math.Min(handStack.StackSize, byPlayer.Entity.Controls.ShiftKey ? 1 : byPlayer.Entity.Controls.CtrlKey ? 32 : 4); //Fixed it so you can actually shift click just 1 in, but also added a ctrl click 32 option just to speed things along.
+
+                    while (desiredTransferAmount * (float)hprops.LitresPerItem + juiceableLitresLeft + juiceableLitresTransfered > juiceableLitresCapacity) desiredTransferAmount -= 1; //Quick calculation of the total number of fruit you can add without going over the total capacity to make sure it's never adding more than a single fruit worth of juice at a time.
```
pizza2004, line 406: Comment was bad.  Meant “to make sure it’s never adding a partial fruit’s worth of juice and not removing the fruit from the hand of the player” haha.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
-                    removeItems = 1;
+                    // the juiceableLitresLeft and juiceableLitresTransfered are per item if we have a stack of multiple
+                    // so we check if the press is full and only remove and add the liters for one item at a time
+                    if (juiceableLitresLeft + juiceableLitresTransfered + transferableLitres + usedLitres > juiceableLitresCapacity)
+                    {
+                        (Api as ICoreClientAPI)?.TriggerIngameError(this, "fullcontainer", Lang.Get("Container is full, press out juice and remove the mash before adding more"));
+                        return false;
+                    }
+
+                    if (juiceableLitresLeft + juiceableLitresTransfered <= 0) mashStack.Attributes.SetDouble("squeezeRel", handStack.Attributes.GetDouble("squeezeRel", 1)); //Sets the mash in the press to be as squeezed as the one you're adding by hand.
```
pizza2004, line 401: This is specifically so that if you’re adding a wet mash to the empty press it will be as squeezed as it was before.  If there’s already something in the press it will reset the squeeze entirely.

BlockEntity/BEFruitpress.cs
```diff
@@ -376,9 +369,11 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
                 var hprops = getJuiceableProps(handStack);
                 if (hprops == null) return false;
 
+                if (hprops.LitresPerItem == null && !handStack.Attributes.HasAttribute("juiceableLitresLeft")) return false; //Prevents dry mash from being inserted.
+
                 var pressedStack = hprops.PressedStack.ResolvedItemstack.Clone();
                 if (MashSlot.Empty) MashSlot.Itemstack = pressedStack;
-                else if (mashStack.StackSize >= 10)
+                else if (juiceableLitresLeft + juiceableLitresTransfered >= juiceableLitresCapacity) //Now properly displays if the container is full once you've maximum fruit.  Includes transfered liters so it's not possible to add more than a stack's worth of fruit slowly by juicing some and then adding more.
```
pizza2004, line 376: This is referencing a thing you could do before of releasing the press, pulling out the mash, adding it back in, and adding more fruit until “juiceableLitersLeft” hit 10 again, which would sometimes mean adding a partial fruit worth of juice, allowing for a very tedious way of getting infinite juice out of just one fruit if we didn’t have these checks in place.

BlockEntity/BEFruitpress.cs
```diff
@@ -500,14 +512,8 @@ public void OnBlockInteractStop(float secondsUsed, IPlayer byPlayer)
 
             if (!CompressAnimActive) return;
 
-            if (secondsUsed >= 4.8f)
```
pizza2004, line 503: I noticed that when I removed this and the earlier seconds used conditions the press started to act more predictably.  The server animation would flip out if I changed it to run at 0.5f instead of 1f before I removed these conditions, but removing them not only fixed that, it also fixed the issue where you could hold the crank down for 4.8 seconds and then let go of the mouse button and it would finish screwing by itself.  Or sometimes just clicking right click fast enough would get it to screw down all by itself from the top.  That felt unintended.

BlockEntity/BEFruitpress.cs
```diff
@@ -269,23 +274,11 @@ private void onTick100msServer(float dt)
...
-            if (juiceableLitresLeft <= 0.01)
+            if (juiceableLitresLeft < 0.01)
             {
-                // Hack to fix rounding errors
-                if (cntBlock != null)
-                {
-                    float litres = cntBlock.GetCurrentLitres(BucketSlot.Itemstack);
-                    cntBlock.SetCurrentLitres(BucketSlot.Itemstack, (float)Math.Round(100 * litres) / 100f);
-                }
-
```
pizza2004, line 280: This hack is no longer necessary since we are simply rounding the value in litersToTransfer to 2 decimal places up above before it moves it around.

BlockEntity/BEFruitpress.cs
```diff
@@ -269,23 +274,11 @@ private void onTick100msServer(float dt)
...
                 UnregisterGameTickListener(listenerId);
                 listenerId = 0;
 
-                int stacksize = GameMath.RoundRandom(Api.World.Rand, (float)juiceableLitresTransfered);
-                mashStack.Attributes.RemoveAttribute("juiceableLitresTransfered");
-                mashStack.Attributes.RemoveAttribute("juiceableLitresLeft");
-                mashStack.StackSize = (int)(stacksize * juiceProps.PressedDryRatio);
-
```
pizza2004, line 288: This is moved to when the player removes the mash from the fruit press.  That makes it so that players can’t accidentally overwrite the dry mash inside with wet mash by adding more fruit, keeping the behavior of mash inside the machine more predictable.

BlockEntity/BEFruitpress.cs
```diff
@@ -93,6 +93,8 @@ BlockEntityAnimationUtil animUtil
 
         long listenerId;
         double juiceableLitresCapacity = 10;
+        double screwPercent; //What percent from the bottom of the animation are we.
```
pizza2004, line 96: What percent from the bottom of the animation are we.

BlockEntity/BEFruitpress.cs
```diff
@@ -376,9 +369,11 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
                 var hprops = getJuiceableProps(handStack);
                 if (hprops == null) return false;
 
+                if (hprops.LitresPerItem == null && !handStack.Attributes.HasAttribute("juiceableLitresLeft")) return false; //Prevents dry mash from being inserted.
```
pizza2004, line 372: Prevents dry mash from being inserted.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
+                        return false;
+                    }
+
+                    if (juiceableLitresLeft + juiceableLitresTransfered <= 0) mashStack.Attributes.SetDouble("squeezeRel", handStack.Attributes.GetDouble("squeezeRel", 1)); //Sets the mash in the press to be as squeezed as the one you're adding by hand.
                 } else
-                {  
-                    float desiredTransferSizeLitres = byPlayer.Entity.Controls.ShiftKey ? (float)hprops.LitresPerItem : Math.Min(handStack.StackSize, 4) * (float)hprops.LitresPerItem;
-                    transferableLitres = (float)Math.Min(desiredTransferSizeLitres, juiceableLitresCapacity - juiceableLitresLeft);
+                {
+                    int desiredTransferAmount = Math.Min(handStack.StackSize, byPlayer.Entity.Controls.ShiftKey ? 1 : byPlayer.Entity.Controls.CtrlKey ? 32 : 4); //Fixed it so you can actually shift click just 1 in, but also added a ctrl click 32 option just to speed things along.
```
pizza2004, line 404: Fixed it so you can actually shift click just 1 in, but also added a ctrl click 32 option just to speed things along.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
+
+                    while (desiredTransferAmount * (float)hprops.LitresPerItem + juiceableLitresLeft + juiceableLitresTransfered > juiceableLitresCapacity) desiredTransferAmount -= 1; //Quick calculation of the total number of fruit you can add without going over the total capacity to make sure it's never adding more than a single fruit worth of juice at a time.
 
-                    removeItems = (int)(transferableLitres / hprops.LitresPerItem);
+                    transferableLitres = desiredTransferAmount * (float)hprops.LitresPerItem;
+                    removeItems = desiredTransferAmount;
                 }
 
-                if (transferableLitres > 0) {
+                if (removeItems > 0) { //Changed to only run this when it is going to remove items instead of just any time it thinks there's juice to add.
```
pizza2004, line 412: Changed to only run this when it is going to remove items instead of just any time it thinks there's juice to add.

BlockEntity/BEFruitpress.cs
```diff
@@ -425,6 +427,16 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
             // Take out mash
             if (MashSlot.Empty) return false;
 
+            if (juiceableLitresLeft < 0.01 && mashStack.Collectible.Code.Path != "rot") // I moved this block of code to when the player removes the mash so you never end up in the situation where the player is trying to add more fruit while there is a stack of finished dry mash in the press. It also removes squeezeRel fixing the stacking bug.
```
pizza2004, line 430: I moved this block of code to when the player removes the mash so you never end up in the situation where the player is trying to add more fruit while there is a stack of finished dry mash in the press. It also removes squeezeRel fixing the stacking bug.

BlockEntity/BEFruitpress.cs
```diff
@@ -519,9 +525,12 @@ private void udpateSqueezeRel(RunningAnimation anim)
             float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
 
             squeezeRel += Math.Max(0, 0.9f - selfHeight);
+            pressSqueezeRel = GameMath.Clamp(squeezeRel, 0.1f, 1f); //Store the value before performing the Min function so we can tell if the screw is touching the mash yet by comparing the two relative squeezes, in case the player releases and cranks back down.
```
pizza2004, line 528: Store the value before performing the Min function so we can tell if the screw is touching the mash yet by comparing the two relative squeezes, in case the player releases and cranks back down.

BlockEntity/BEFruitpress.cs
```diff
@@ -519,9 +525,12 @@ private void udpateSqueezeRel(RunningAnimation anim)
             float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
 
             squeezeRel += Math.Max(0, 0.9f - selfHeight);
+            pressSqueezeRel = GameMath.Clamp(squeezeRel, 0.1f, 1f); //Store the value before performing the Min function so we can tell if the screw is touching the mash yet by comparing the two relative squeezes, in case the player releases and cranks back down.
             squeezeRel = GameMath.Clamp(Math.Min(mashStack.Attributes.GetDouble("squeezeRel", 1), squeezeRel), 0.1f, 1f);
 
             mashStack.Attributes.SetDouble("squeezeRel", squeezeRel);
+
+            screwPercent = GameMath.Clamp(1f - anim.CurrentFrame / (anim.Animation.QuantityFrames - 1), 0, 1f) / selfHeight; //A quick calculation of how much the mash is squeezed.  Dividing by selfHeight makes it so smaller mash responds more like expected while still always reaching 0% at the end.
```
pizza2004, line 533: A quick calculation of how much the mash is squeezed.  Dividing by selfHeight makes it so smaller mash responds more like expected while still always reaching 0% at the end.

BlockEntity/BEFruitpress.cs
```diff
@@ -180,8 +182,9 @@ public override void Initialize(ICoreAPI api)
         private void onTick25msClient(float dt)
         {
             double squeezeRel = mashStack?.Attributes.GetDouble("squeezeRel", 1) ?? 1;
+            float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f; //Added so the particles will better match the actual position of the mash.
```
pizza2004, line 185: Added so the particles will better match the actual position of the mash.

BlockEntity/BEFruitpress.cs
```diff
@@ -230,17 +233,19 @@ private void onTick100msServer(float dt)
             double totalHours = Api.World.Calendar.TotalHours;
 
             double squeezeRel = mashStack.Attributes.GetDouble("squeezeRel", 1);
-            double litresToTransfer = Math.Min(juiceableLitresLeft, (totalHours - lastLiquidTransferTotalHours) * 50f);
+            double litresJuiceable = Math.Max(0, juiceableLitresLeft - ((juiceableLitresLeft + juiceableLitresTransfered) * screwPercent)); //Determines how much of the mash we can juice for how far down the screw is cranked.
```
pizza2004, line 236: Determines how much of the mash we can juice for how far down the screw is cranked.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
+                    {
+                        (Api as ICoreClientAPI)?.TriggerIngameError(this, "fullcontainer", Lang.Get("Container is full, press out juice and remove the mash before adding more"));
+                        return false;
+                    }
+
+                    if (juiceableLitresLeft + juiceableLitresTransfered <= 0) mashStack.Attributes.SetDouble("squeezeRel", handStack.Attributes.GetDouble("squeezeRel", 1));
                 } else
-                {  
-                    float desiredTransferSizeLitres = byPlayer.Entity.Controls.ShiftKey ? (float)hprops.LitresPerItem : Math.Min(handStack.StackSize, 4) * (float)hprops.LitresPerItem;
-                    transferableLitres = (float)Math.Min(desiredTransferSizeLitres, juiceableLitresCapacity - juiceableLitresLeft);
```
pizza2004, line 406: Quick calculation of the total number of fruit you can add without going over the total capacity to make sure it’s never adding a partial fruit’s worth of juice and not removing the fruit from the hand of the player.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,28 +385,35 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
-                    removeItems = 1;
+                    // the juiceableLitresLeft and juiceableLitresTransfered are per item if we have a stack of multiple
+                    // so we check if the press is full and only remove and add the liters for one item at a time
+                    if (juiceableLitresLeft + juiceableLitresTransfered + transferableLitres + usedLitres > juiceableLitresCapacity)
+                    {
+                        (Api as ICoreClientAPI)?.TriggerIngameError(this, "fullcontainer", Lang.Get("Container is full, press out juice and remove the mash before adding more"));
+                        return false;
+                    }
+
+                    if (juiceableLitresLeft + juiceableLitresTransfered <= 0) mashStack.Attributes.SetDouble("squeezeRel", handStack.Attributes.GetDouble("squeezeRel", 1));
```
pizza2004, line 401: Sets the mash in the press to be as squeezed as the one you're adding by hand. This is specifically so that if you’re adding a wet mash to the empty press it will be as squeezed as it was before. If there’s already something in the press it will reset the squeeze entirely.

BlockEntity/BEFruitpress.cs
```diff
@@ -376,9 +369,11 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
                 var hprops = getJuiceableProps(handStack);
                 if (hprops == null) return false;
 
+                if (hprops.LitresPerItem == null && !handStack.Attributes.HasAttribute("juiceableLitresLeft")) return false;
+
                 var pressedStack = hprops.PressedStack.ResolvedItemstack.Clone();
                 if (MashSlot.Empty) MashSlot.Itemstack = pressedStack;
-                else if (mashStack.StackSize >= 10)
+                else if (juiceableLitresLeft + juiceableLitresTransfered >= juiceableLitresCapacity)
```
pizza2004, line 376: Now properly displays if the container is full once you've hit juiceableLitresCapacity. Includes transferred liters so it's not possible to add more than a stack's worth of fruit slowly by juicing some and then adding more. Before you could release the press, pull out the mash, insert it back in, and add more fruit until “juiceableLitersLeft” hit 10 again, which would sometimes mean adding a partial fruit worth of juice, allowing for a very tedious way of getting infinite juice out of just one fruit if we didn’t have these checks in place.

BlockEntity/BEFruitpress.cs
```diff
@@ -93,6 +93,8 @@ BlockEntityAnimationUtil animUtil
 
         long listenerId;
         double juiceableLitresCapacity = 10;
+        double screwPercent;
```
pizza2004, line 96: What percent from the bottom of the animation are we.

BlockEntity/BEFruitpress.cs
```diff
@@ -93,6 +93,8 @@ BlockEntityAnimationUtil animUtil
 
         long listenerId;
         double juiceableLitresCapacity = 10;
+        double screwPercent;
+        double pressSqueezeRel;
```
pizza2004, line 97: An internal version of squeezeRel for comparison to make sure the crank is touching the mash before it activates.

BlockEntity/BEFruitpress.cs
```diff
@@ -230,17 +233,19 @@ private void onTick100msServer(float dt)
             double totalHours = Api.World.Calendar.TotalHours;
 
             double squeezeRel = mashStack.Attributes.GetDouble("squeezeRel", 1);
-            double litresToTransfer = Math.Min(juiceableLitresLeft, (totalHours - lastLiquidTransferTotalHours) * 50f);
+            double litresJuiceable = Math.Max(0, juiceableLitresLeft - ((juiceableLitresLeft + juiceableLitresTransfered) * screwPercent));
+
+            double litresToTransfer = Math.Min(litresJuiceable, (float)Math.Round((totalHours - lastLiquidTransferTotalHours) * 50f, 2));
```
pizza2004, line 238: Rounding to two decimal points here removes the need for the rounding hack later.

BlockEntity/BEFruitpress.cs
```diff
@@ -376,9 +369,11 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
                 var hprops = getJuiceableProps(handStack);
                 if (hprops == null) return false;
 
+                if (hprops.LitresPerItem == null && !handStack.Attributes.HasAttribute("juiceableLitresLeft")) return false;
```
pizza2004, line 372: Prevents dry mash from being inserted.

BlockEntity/BEFruitpress.cs
```diff
@@ -425,6 +427,16 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
             // Take out mash
             if (MashSlot.Empty) return false;
 
+            if (juiceableLitresLeft < 0.01 && mashStack.Collectible.Code.Path != "rot")
```
pizza2004, line 430: I moved this block of code to when the player removes the mash so you never end up in the situation where the player is trying to add more fruit while there is a stack of finished dry mash in the press. It also removes squeezeRel fixing the stacking bug.

BlockEntity/BEFruitpress.cs
```diff
@@ -519,9 +525,12 @@ private void udpateSqueezeRel(RunningAnimation anim)
             float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
 
             squeezeRel += Math.Max(0, 0.9f - selfHeight);
+            pressSqueezeRel = GameMath.Clamp(squeezeRel, 0.1f, 1f);
```
pizza2004, line 528: Store the value before performing the Min function so we can tell if the screw is touching the mash yet by comparing the two relative squeezes, in case the player releases and cranks back down.

BlockEntity/BEFruitpress.cs
```diff
@@ -519,9 +525,12 @@ private void udpateSqueezeRel(RunningAnimation anim)
             float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
 
             squeezeRel += Math.Max(0, 0.9f - selfHeight);
+            pressSqueezeRel = GameMath.Clamp(squeezeRel, 0.1f, 1f);
             squeezeRel = GameMath.Clamp(Math.Min(mashStack.Attributes.GetDouble("squeezeRel", 1), squeezeRel), 0.1f, 1f);
 
             mashStack.Attributes.SetDouble("squeezeRel", squeezeRel);
+
+            screwPercent = GameMath.Clamp(1f - anim.CurrentFrame / (anim.Animation.QuantityFrames - 1), 0, 1f) / selfHeight;
```
pizza2004, line 533: A quick calculation of how much the mash is squeezed. Dividing by selfHeight makes it so smaller mash responds more like expected while still always reaching 0% at the end.

BlockEntity/BEFruitpress.cs
```diff
@@ -180,8 +182,9 @@ public override void Initialize(ICoreAPI api)
         private void onTick25msClient(float dt)
         {
             double squeezeRel = mashStack?.Attributes.GetDouble("squeezeRel", 1) ?? 1;
+            float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
```
pizza2004, line 185: Added so the particles will better match the actual position of the mash.

BlockEntity/BEFruitpress.cs
```diff
@@ -230,17 +233,19 @@ private void onTick100msServer(float dt)
             double totalHours = Api.World.Calendar.TotalHours;
 
             double squeezeRel = mashStack.Attributes.GetDouble("squeezeRel", 1);
-            double litresToTransfer = Math.Min(juiceableLitresLeft, (totalHours - lastLiquidTransferTotalHours) * 50f);
+            double litresJuiceable = Math.Max(0, juiceableLitresLeft - ((juiceableLitresLeft + juiceableLitresTransfered) * screwPercent));
```
pizza2004, line 236: Determines how much of the mash we can juice for how far down the screw is cranked.

pizza2004:
My latest commits fix it so that breaking the block properly produces the same expected amount of attribute cleaned dry mash as removing it by hand would, as well as making wet pressed mash produce the same amount of rot as it would have if you had fully juiced it first.

BlockEntity/BEFruitpress.cs
```diff
@@ -497,7 +489,7 @@ public bool OnBlockInteractStep(float secondsUsed, IPlayer byPlayer, EnumFruitPr
                 RunningAnimation anim = animUtil.animator.GetAnimationState("compress");
                 if (anim != null)
                 {
-                    udpateSqueezeRel(anim);
+                    updateSqueezeRel(anim);
```
Craluminum2413, line 492: Use this since updateSqueezeRel checks for null already
```cs
updateSqueezeRel(animUtil.animator.GetAnimationState("compress"));
```

pizza2004: Oh, you're right, thanks!

pizza2004:
I have submitted a couple more commits.  They fix a bunch of little issues I couldn't figure out before, which should make the whole experience more consistent.  For instance, it should keep the server and client even more properly in-sync.  Furthermore I have slightly altered the logic so that mash will drip a little extra out slowly if you unscrew the press before it is fully dry, rather than simply immediately stopping.

And I fixed a bug with the renderer in regards to Rot.  I had just assumed it was doing it because I was using different classes for such a long time I didn't even realize it was broken.  This should fix the visual glitch entirely.  I also made it so that when you add a pressed mash to the machine it will merge the transition states if there is already one inside, and if it's empty it will simply place it directly in rather than needing to manipulate it.

BlockEntity/BEFruitpress.cs
```diff
@@ -198,7 +204,7 @@ private void onTick25msClient(float dt)
 
                 minPos.Mul(8 / 16f);
                 maxPos.Mul(8 / 16f);
```
pizza2004, line 206: Test 1

BlockEntity/BEFruitpress.cs
```diff
@@ -198,7 +204,7 @@ private void onTick25msClient(float dt)
 
                 minPos.Mul(8 / 16f);
                 maxPos.Mul(8 / 16f);
-                maxPos.Y = 5 / 16f - (1 - squeezeRel) * 0.25f;
+                maxPos.Y = 5 / 16f - (1 - squeezeRel + Math.Max(0, 0.9f - selfHeight)) * 0.5f;
```
pizza2004, line 208: Test 2

pizza2004 (commented):
Test 3

BlockEntity/BEFruitpress.cs
```diff
@@ -180,8 +186,9 @@ public override void Initialize(ICoreAPI api)
         private void onTick25msClient(float dt)
         {
             double squeezeRel = mashStack?.Attributes.GetDouble("squeezeRel", 1) ?? 1;
+            float selfHeight = (float)(juiceableLitresTransfered + juiceableLitresLeft) / 10f;
```
pizza2004, line 189: Added so the particles will better match the actual position of the mash.

BlockEntity/BEFruitpress.cs
```diff
@@ -263,29 +276,17 @@ private void onTick100msServer(float dt)
...
-
+                squeezedJuiceLeft = false;
                 UnregisterGameTickListener(listenerId);
                 listenerId = 0;
 
-                int stacksize = GameMath.RoundRandom(Api.World.Rand, (float)juiceableLitresTransfered);
-                mashStack.Attributes.RemoveAttribute("juiceableLitresTransfered");
-                mashStack.Attributes.RemoveAttribute("juiceableLitresLeft");
-                mashStack.StackSize = (int)(stacksize * juiceProps.PressedDryRatio);
-
```
pizza2004, line 288: This is moved into a function that is run when the player removes the mash from the fruit press or the press is broken. That makes it so that players can’t accidentally overwrite the dry mash inside with wet mash by adding more fruit, keeping the behavior of mash inside the machine more predictable.

BlockEntity/BEFruitpress.cs
```diff
@@ -390,25 +418,59 @@ private bool InteractMashContainer(IPlayer byPlayer, BlockSelection blockSel)
...
+
+                    removeItems =  1;
                 } else
                 {
-                    float desiredTransferSizeLitres = byPlayer.Entity.Controls.ShiftKey ? (float)hprops.LitresPerItem : Math.Min(handStack.StackSize, 4) * (float)hprops.LitresPerItem;
-                    transferableLitres = (float)Math.Min(desiredTransferSizeLitres, juiceableLitresCapacity - juiceableLitresLeft);
+                    // In order to make sure we're always giving exactly the amount of juice that's appropriate for the
+                    // number of items we take out of the inventory we start by counting up how many the player wants to
+                    // add, be it 1, 4, or 32, and then subtracting 1 from that total until we have an amount that can fit
+                    int desiredTransferAmount = Math.Min(handStack.StackSize, byPlayer.Entity.Controls.ShiftKey ? 1 : byPlayer.Entity.Controls.CtrlKey ? 32 : 4);
```
pizza2004, line 458: The change to BlockFruitpressTop fixes this so you can actually Shift + Click just 1 in.  I also added a Ctrl + Click 32 option just to speed things along.

BlockEntity/BEFruitpress.cs
```diff
@@ -505,55 +575,58 @@ public bool OnBlockInteractStep(float secondsUsed, IPlayer byPlayer, EnumFruitPr
...
 
         public void OnBlockInteractStop(float secondsUsed, IPlayer byPlayer)
         {
-            RunningAnimation anim = animUtil.animator.GetAnimationState("compress");
-            udpateSqueezeRel(anim);
+            updateSqueezeRel(animUtil.animator.GetAnimationState("compress"));
 
             if (!CompressAnimActive) return;
 
-            if (secondsUsed >= 4.8f)
```
pizza2004, line 528: I noticed that when I removed this and the earlier seconds used conditions the press started to act more predictably. The server animation would flip out if I changed it to run at 0.5f instead of 1f before I removed these conditions, but removing them not only fixed that, it also fixed the issue where you could hold the crank down for 4.8 seconds and then let go of the mouse button, or sometimes just clicking right click fast enough, and it would finish screwing by itself.

BlockEntity/BEFruitpress.cs
```diff
@@ -601,11 +675,6 @@ public override void OnReceivedServerPacket(int packetid, byte[] data)
 
                 if (packet.AnimationActive)
                 {
-                    if (!MashSlot.Empty && juiceableLitresLeft > 0 && !CompressAnimActive)
```
pizza2004, line 604: Since the server tells all clients to play the sound at a different point in time _without_ this bit it felt unnecessary and it also meant the sound played at a different time for people who weren’t the ones using the machine.

Item/ItemPressedMash.cs
```diff
@@ -15,6 +16,14 @@ public override string GetHeldItemName(ItemStack itemStack)
             return Lang.GetMatching(Code?.Domain + AssetLocation.LocationSeparator + type + "-" + Code?.Path + "-" + ap);
         }
 
+        public override ItemStack OnTransitionNow(ItemSlot slot, TransitionableProperties props)
```
pizza2004, line 19: Adding this to the PressedMash item class insures that you’ll always get the appropriate amount of rot out, the same as you would if you had finished pressing it.  Since we are no longer converting the mash before it is removed from the press it helps make sure the right amount ends up in the machine if it rots as well.

BlockEntityRenderer/FruitpressContentsRenderer.cs
```diff
@@ -98,9 +96,10 @@ public void reloadMeshes(JuiceableProperties props, bool mustReload)
             int y;
             if (stack.Collectible.Code.Path == "rot")
             {
-                textureLocation = new AssetLocation("block/rot/rot");
-                y = GameMath.Clamp(stack.StackSize / 4, 1, 9);
-            } else
+                textureLocation = new AssetLocation("block/wood/barrel/rot");
+                y = GameMath.Clamp(stack.StackSize / 2, 1, 9);
```
pizza2004, line 100: This looks better visually / 2 in my opinion.  Since the stack can never be bigger than 10 it ends up returning only 2 or 1, meaning it’s less obvious how much rot is inside comparatively, with / 2 it can range from 5 down to 1, meaning it still ends up not being full, but still having enough to visually parse how much material is inside.

pizza2004 (commented):
Various comments added to explain the reasoning for while things were changed or what they are doing, but that aren’t relevant for when you are just reading the code normally.

BlockEntity/BEFruitpress.cs
```diff
@@ -230,17 +237,23 @@ private void onTick100msServer(float dt)
             double totalHours = Api.World.Calendar.TotalHours;
 
             double squeezeRel = mashStack.Attributes.GetDouble("squeezeRel", 1);
-            double litresToTransfer = Math.Min(juiceableLitresLeft, (totalHours - lastLiquidTransferTotalHours) * 50f);
 
-            if (Api.Side == EnumAppSide.Server && squeezeRel < 1 && !squeezeSoundPlayed && juiceableLitresLeft > 0)
+            // First we need to calculate how squeezed down the mash has been and therefore how much we're allowed to take
+            squeezedLitresLeft = Math.Max(Math.Max(0, squeezedLitresLeft), juiceableLitresLeft - ((juiceableLitresLeft + juiceableLitresTransfered) * screwPercent));
+            double litresToTransfer = Math.Min(squeezedLitresLeft, Math.Round((totalHours - lastLiquidTransferTotalHours) * (CompressAnimActive ? GameMath.Clamp(squeezedLitresLeft * (1 - squeezeRel) * 500f, 25f, 100f) : 5f), 2));
```
pizza2004, line 243: Rounding this to two decimal places removes the need for the “hack” from later in the code.

pizza2004:
My latest commit fixes it so that the animation state of the screw is properly saved and loaded to the TreeAttributes and sychronized over multiplayer so there should never again be the issue where if someone logs out and in the screw will reset to the top.

tyronx:
I have merged this now, thanks a lot!

PR #63, opened 2024-09-19
Fix parameter name typo, restore API surface

dmchurch:
The real fix here is in lines 473/474, fixing a variable name collision issue. The extra parameterless method is to re-add the public API method you removed and which I was, in fact, using in a mod (which is how I found the naming issue in the first place - my mod simply stopped linking against the 1.20 DLLs).

### DIFF
--- a/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs
+++ b/Systems/WorldGen/Standard/ChunkGen/5.GenDeposits/Generators/DiscGenerator.cs
@@ -201,8 +201,7 @@ namespace Vintagestory.ServerMods
                 Api.Server.LogWarning("Deposit in file {0} has no inblock defined, it will never spawn.", variant.fromFile);
             }
 
-            var rnd = new LCGRandom(Api.World.Seed);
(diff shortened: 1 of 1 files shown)


Th3Dilli:
oh thanks good find

but the point of needing to provide an instance of LCGRandom is that when you call the method again it does not produce the same output again, if we re-add the parameter less version and you call that on a every region/chunk etc it will always produce the same result if `Radius` and `Thickness` are unchanged which is the default behavior

we had this issue with some systems like blockpatches and others where LCGRandom was not used or not properly used, with 1.20 this should be mostly solved and running `/wgen regen` should produce to 99% the same result of the chunk

if only need it to run once i would recommend to just do as we do with setting up a new LCGRandom and call it once and cache the value and you should be good

so I am not that sold on the parameter less version of it

Th3Dilli:
I have now made it public and made use of the LCGRandom I hope that works for you

Th3Dilli:
ok sorry for the inconvenience, I have now restored the old method for API compatibility

dmchurch:
Thank you very much 🙂

PR #66, opened 2024-09-22
Use CodeWithVariant and Variant where it is possible to avoid unpredictable behavior

Craluminum2413:
When I added wooden variants to all mechanical blocks, they were crashing because they didn't expect new variants.

These fixes allow variants for mechanical blocks and crucibles.
The rest of fixes may not be that important, but they worth adding too.


pizza2004:
This will need to be done in smaller chunks that are easier to test.  I will close this, but thank you for the submission.

PR #68, opened 2024-09-25
Avoid removing of bed when unmounting

Craluminum2413:
It shouldn't call OnBlockRemoved and instead should call unregister tick listener


tyronx:
wow strange. Thanks, now merged.

PR #69, opened 2024-09-25
Fix barrel sounds

Craluminum2413:
This simple fix allows to set custom sound for barrel through attributes.

When modding barrels, they expect existing sound in `modid;sounds/block/barrel*`, thus 100 mods that add barrels require 200 new sounds, even if they want to use default sounds.


pizza2004:
Thanks, this will be in a future release.

PR #70, opened 2024-09-25
Allow custom drops for basket trap

Craluminum2413:
Make it moddable


pizza2004:
We will be handling this a different way, but thank you for your submission!

PR #71, opened 2024-09-25
Make firepit heat clay oven above

Craluminum2413:


pizza2004:
This seems to me better suited to a mod or developed in a more purposeful way with new blocks, but thank you for the code submission!

PR #72, opened 2024-09-25
Show total amount of metal units by calculating them together with stack size

Craluminum2413:
- Show total amount of metal units using stack size


pizza2004:
This seems unnecessary and possibly even confusing to me and I think it is best left to a mod, but thank you for the submission.

PR #73, opened 2024-09-26
Fix crash related to watering can when loading world

Craluminum2413:
Fixes this issue: https://github.com/anegostudios/VintageStory-Issues/issues/2829


Craluminum2413:
#2829

tyronx:
the team already fixed this, but thanks

PR #74, opened 2024-09-29
Use GetCollectibleInterface/GetInterface where possible

Craluminum2413:
Use new CollectibleObject.GetCollectibleInterface 1.20 method where possible.
Also use Block.GetInterface where possible too.

These two methods make the game more moddable, so they should be used everywhere


Craluminum2413:
I did it for every single class in the game

Craluminum2413:
I don't know what I just did

pizza2004:
This is very complicated as a PR and has a lot of issues, so I am closing it in favor of things being done in a more piecemeal fashion.

PR #75, opened 2024-09-29
Allow randomizers to be ground storable

Craluminum2413:
For randomized items in ground storage I guess


tyronx:
whats the reason for this?

Craluminum2413:
@tyronx To randomize ground storable items, for example, you place "shield randomizer" and it places random shield, or you place "tool randomizer" and it places random tool there against wall

tyronx:
that would require a more code. 
- You can't Shift right click them on the ground, because that opens the GUI 
- it would need to delete the ground storage block if the randomizer resolves to nothing

pizza2004:
Will you be fixing the issues Tyron brought up?  If not, please close this.

PR #76, opened 2024-10-03
Overhaul Charcoal Pit and Charcoal Pile

pizza2004:
The old code would check ± 6 in each direction, which meant that it was actually checking a 13x13x13 cube with the fire pit at the center.  I have slightly modified it so it will check an 11x11x11 cube located _under_ the fire pit, with the fire pit in the center.  This should match the handbook saying that “your charcoal pit can be any dimension up to 11x11x11 blocks” and “Cover the filled pit completely, but leave the top of a single stack of firewood in the center exposed to air.”

It might be desirable to make it so it doesn’t have to be centered in the cube horizontally, but that would take a more complex fix, while this should make it match the explanation from the handbook.

Edit: With the Overhaul Charcoal Pile commit the JSON needs to be updated to add `fullBlockCode: "charcoalpile-8"` to the attributes as well as the UnstableFalling block behavior.  More in the later comments.

### DIFF
--- a/Block/BlockCharcoalPile.cs
+++ b/Block/BlockCharcoalPile.cs
@@ -1,9 +1,66 @@
 ﻿using Vintagestory.API.Common;
+using Vintagestory.API.Datastructures;
+using Vintagestory.API.MathTools;
 
 namespace Vintagestory.GameContent
 {
     public class BlockCharcoalPile : BlockLayeredSlowDig
     {
+        public override bool OnFallOnto(IWorldAccessor world, BlockPos pos, Block block, TreeAttribute blockEntityAttributes)
+        {
+            Block nBlock = (BlockCharcoalPile)world.BlockAccessor.GetMostSolidBlock(pos);
+            if (block is BlockCharcoalPile && nBlock is BlockCharcoalPile)
+            {
+
+                Block uBlock = block;
+                while (((BlockCharcoalPile)nBlock).CountLayers() < 8 && uBlock != null)
+                {
+                    nBlock = ((BlockCharcoalPile)nBlock).GetNextLayer(world);
+                    uBlock = ((BlockCharcoalPile)uBlock).GetPrevLayer(world);
+                }
+
+                world.BlockAccessor.SetBlock(nBlock.BlockId, pos);
+                world.BlockAccessor.TriggerNeighbourBlockUpdate(pos);
+                if (uBlock != null)
+                {
+                    world.BlockAccessor.SetBlock(uBlock.BlockId, pos.UpCopy());
+                    world.BlockAccessor.TriggerNeighbourBlockUpdate(pos.UpCopy());
+                }
+                return true;
+            }
+
+            return base.OnFallOnto(world, pos, block, blockEntityAttributes);
+        }
+
+        public override void OnBlockBroken(IWorldAccessor world, BlockPos pos, IPlayer byPlayer, float dropQuantityMultiplier = 1f)
+        {
+            Block block = world.BlockAccessor.GetBlock(pos);
+
+            BlockPos abovePos = pos.UpCopy();
+            Block aboveBlock = null;
+
+            while (abovePos.Y < world.BlockAccessor.MapSizeY)
+            {
+                aboveBlock = world.BlockAccessor.GetBlock(abovePos);
+                if (aboveBlock.FirstCodePart() != block.FirstCodePart()) break;
+
+                abovePos.Up();
+            }
+
+            if (abovePos == pos.UpCopy() || aboveBlock == null || byPlayer?.WorldData.CurrentGameMode == EnumGameMode.Creative)
+            {
+                base.OnBlockBroken(world, pos, byPlayer);
+                world.BlockAccessor.TriggerNeighbourBlockUpdate(pos);
+                return;
+            }
+            else if (aboveBlock.FirstCodePart() == block.FirstCodePart()) aboveBlock.OnBlockBroken(world, abovePos, byPlayer);
+            else
+            {
+                BlockPos topPos = abovePos.DownCopy();
+                Block topBlock = world.BlockAccessor.GetBlock(topPos);
+                topBlock.OnBlockBroken(world, topPos, byPlayer);
+            }
+        }
 
         public override float RandomSoundPitch(IWorldAccessor world)
         {
--- a/BlockEntity/BECharcoalPit.cs
+++ b/BlockEntity/BECharcoalPit.cs
@@ -31,6 +31,10 @@ namespace Vintagestory.GameContent
 
         public bool Lit => lit;
 
+        int maxSize = 11;
+
+        public int MaxPileSize { get { return maxSize; } set { maxSize = value; } }
+
 
         public override void Initialize(ICoreAPI api)
         {
@@ -47,7 +51,7 @@ namespace Vintagestory.GameContent
 
             if (Lit)
             {
-                FindHoleInPit();
+                FindHolesInPit();
             }
         }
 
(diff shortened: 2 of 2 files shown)


pizza2004:
I have updated it slightly so that it is possible to set the MaxPileSize publicly now, allowing modding to change how big you can make charcoal pits more easily.

pizza2004:
I have now edited it so that it will check for any configuration of firewood with a firepit sitting on top of a stack that fits inside of a maxSize range cube in all 3 directions.

pizza2004:
I sat down to really test out everything and realized there were more improvements to be made.  In some instances I fixed vanilla issues (not using .Copy() on BlockPos was causing unintended behavior), but mostly what I've done here is remove repetitive code and make the Charcoal Pit aware of the flammability of the blocks making up its walls.  This makes it so that any combustible blocks count as "holes" and they can light on fire if exposed to air, as well as just generally making the pit not work.

It also makes it so the method returns _all_ the holes in the pit and they are all lit on fire.  Perhaps it could be altered so that it would not light them all at the same moment, but before it would not light any but the first it found, and even then only if there was air allowing that one to be lit.

pizza2004:
I have now overhauled the charcoal pile block itself somewhat, applying gravity so that it will not float in the air, and adding code so that the blocks will merge as they fall to prevent gaps from forming between them.  I also made it so that if you are shoveling the charcoal it will break the block at the top of the stack instead of the one you are directly pointing at, so that stacks will break top down the way you would expect.

Block/BlockCharcoalPile.cs
```diff
@@ -1,9 +1,66 @@
...
+                if (uBlock != null)
+                {
+                    world.BlockAccessor.SetBlock(uBlock.BlockId, pos.UpCopy());
+                    world.BlockAccessor.TriggerNeighbourBlockUpdate(pos.UpCopy());
+                }
+                return true;
+            }
+
+            return base.OnFallOnto(world, pos, block, blockEntityAttributes);
+        }
```
pizza2004, line 33: Allowing for falling charcoal piles to merge into the piles they land on when given the gravity behavior.

Block/BlockCharcoalPile.cs
```diff
@@ -1,9 +1,66 @@
...
+                return;
+            }
+            else if (aboveBlock.FirstCodePart() == block.FirstCodePart()) aboveBlock.OnBlockBroken(world, abovePos, byPlayer);
+            else
+            {
+                BlockPos topPos = abovePos.DownCopy();
+                Block topBlock = world.BlockAccessor.GetBlock(topPos);
+                topBlock.OnBlockBroken(world, topPos, byPlayer);
+            }
+        }
```
pizza2004, line 63: This makes it so that when you break the charcoal pile it will remove a layer from the highest block in the overall stack of blocks, to effectively prevent blocks from ending up “floating” up in the air, allowing to more easily dig them up from the bottom.

BlockEntity/BECharcoalPit.cs
```diff
@@ -31,6 +31,10 @@ public class BlockEntityCharcoalPit : BlockEntity
 
         public bool Lit => lit;
 
+        int maxSize = 11;
+
+        public int MaxPileSize { get { return maxSize; } set { maxSize = value; } }
+
```
pizza2004, line 37: Simple way for mods to increase the size of a charcoal pit if desired.

BlockEntity/BECharcoalPit.cs
```diff
@@ -86,38 +90,33 @@ private void OnServerTick(float dt)
...
                 }
 
-                if (firefacing != null)
-                {
-                    Block fireblock = Api.World.GetBlock(new AssetLocation("fire"));
-                    Api.World.BlockAccessor.SetBlock(fireblock.BlockId, holePos);
-                    BlockEntity befire = Api.World.BlockAccessor.GetBlockEntity(holePos);
-                    befire?.GetBehavior<BEBehaviorBurning>()?.OnFirePlaced(firefacing, startedByPlayerUid);
-                }
-
```
pizza2004, line 120: This now looks for every open hole and semi-randomly ignites them over time, keeping it from getting stuck without an obvious fire lighting up under specific circumstances.

BlockEntity/BECharcoalPit.cs
```diff
@@ -203,7 +200,7 @@ void ConvertPit()
                 int maxY = val.Value.Z;
                 while (lpos.Y <= maxY)
                 {
-                    if (BlockFirepit.IsFirewoodPile(Api.World, lpos))  // test for the possibility someone had contiguous firewood both above and below a soil block for example
+                    if (BlockFirepit.IsFirewoodPile(Api.World, lpos) || lpos == Pos)  // test for the possibility someone had contiguous firewood both above and below a soil block for example
```
pizza2004, line 203: The addition of lpos == Pos here removes the need for the SetBlock(0, Pos) lower down while also making sure that the firepit block doesn’t end up being a pocket of air in the middle of a column of charcoal if it was buried in the middle of the firewood.

BlockEntity/BECharcoalPit.cs
```diff
@@ -255,35 +252,70 @@ BlockPos FindHoleInPit()
...
                 {
-                    BlockPos npos = bpos.AddCopy(facing);
+                    facing.IterateThruFacingOffsets(npos);  // This must be the first command in the loop, to ensure all facings will be properly looped through regardless of any 'continue;' statements
                     IWorldChunk chunk = Api.World.BlockAccessor.GetChunkAtBlockPos(npos);
                     if (chunk == null) return null;
 
                     Block nBlock = chunk.GetLocalBlockAtBlockPos(Api.World, npos);
 
-                    bool solid = nBlock.SideSolid[facing.Opposite.Index] || (nBlock is BlockMicroBlock && (chunk.GetLocalBlockEntityAtBlockPos(npos) as BlockEntityMicroBlock).sideAlmostSolid[facing.Opposite.Index]);
+                    bool solid = nBlock.GetLiquidBarrierHeightOnSide(facing.Opposite, npos) == 1 || nBlock.GetLiquidBarrierHeightOnSide(facing, bpos) == 1;
```
pizza2004, line 261: This allows certain things like Iron Hatch Doors, Iron Trapdoors, and Beehive Kiln Doors which don’t have SideSolid set because they shouldn’t allow for things like torches to be placed on them to still count as a solid wall for the sake of a charcoal pit.  Not Iron Doors, though, since they are hard coded to have a window in the top half that liquid can flow through.

BlockEntity/BECharcoalPit.cs
```diff
@@ -255,35 +252,70 @@ BlockPos FindHoleInPit()
...
                     bool isFirewoodpile = BlockFirepit.IsFirewoodPile(Api.World, npos);
 
-                    if (!solid && !isFirewoodpile && nBlock.BlockId != charcoalPitBlockId)
+                    if (!isFirewoodpile && nBlock.BlockId != charcoalPitBlockId)
                     {
-                        return bpos;
+                        if (IsCombustible(npos)) holes.Add(npos.Copy());
+                        else if (!solid) holes.Add(bpos.Copy());
```
pizza2004, line 267: This change allows us to quantify every position where some type of flammable block is exposed to the outside of the charcoal pit, including those making up the walls, rather than just the firewood.

BlockEntity/BECharcoalPit.cs
```diff
@@ -255,35 +252,70 @@ BlockPos FindHoleInPit()
...
+            // Only traverse within maxSize range
+            if (nmax.X - nmin.X + 1 <= maxSize && nmax.Y - nmin.Y + 1 <= maxSize && nmax.Z - nmin.Z + 1 <= maxSize)
+            {
+                minPos = nmin.Copy();
+                maxPos = nmax.Copy();
+                return true;
+            }
+
+            return false;
+        }
```
pizza2004, line 306: Properly updated to construct a pit of the maximum possible size like the room code does where the firepit can be in any block location instead of needing to be in the middle when it gets bigger than a certain size.

BlockEntity/BECharcoalPit.cs
```diff
@@ -255,35 +252,70 @@ BlockPos FindHoleInPit()
...
+        {
+            Block block = Api.World.BlockAccessor.GetBlock(pos);
+            if (block.CombustibleProps != null) return block.CombustibleProps.BurnDuration > 0;
+
+            if (block is ICombustible bic)
+            {
+                return bic.GetBurnDuration(Api.World, pos) > 0;
+            }
+
+            return false;
```
pizza2004, line 318: This function is just used by FindHolesInPit to check whether or not the walls that make up the pit are flammable.  That way even though something like a solid oak door _would_ count as a solid wall, it will light on fire and still not work effectively for the charcoal pit, necessitating that all walls of the pit be built out of properly fire resistant materials.

pizza2004 (commented):
Various explanations for all the changes so the code is easier to read.

tyronx:
this is now merged, thank you

PR #77, opened 2024-10-03
Don't hardcode barrel shapes

Craluminum2413:


Guimoute:
It's bad practice to copy paste

image

5 times. You can clearly see the pattern to write the static stuff only once.

Craluminum2413:
> It's bad practice to copy paste
> 
> image
> 
> 5 times. You can clearly see the pattern to write the static stuff only once.

Do better then

Craluminum2413:
@Guimoute do better then

Systems/Barrel/BlockBarrel.cs
```diff
@@ -345,7 +350,31 @@ public override void OnLoaded(ICoreAPI api)
...
+            }
+            
+            if (Attributes?["contentsShape"].Exists == true)
+            {
+                contentsShape = AssetLocation.Create(Attributes["contentsShape"].AsString("shapes/block/wood/barrel/contents.json")).WithPathPrefixOnce("shapes").WithPathAppendixOnce(".json");
+            }
+                        
+            if (Attributes?["opaqueLiquidContentsShape"].Exists == true)
+            {
+                contentsShape = AssetLocation.Create(Attributes["opaqueLiquidContentsShape"].AsString("shapes/block/wood/barrel/opaqueliquidcontents.json")).WithPathPrefixOnce("shapes").WithPathAppendixOnce(".json");
```
tyronx, line 370: wrong field name here, and this code is indeed a bit suboptimal. will fix.

tyronx: domain field also not correctly set

tyronx: Correct solution
image

tyronx:
this is now merged

PR #83, opened 2024-11-18
Fix Waterfall Not Updating Flow When Blocked on Bottom Side

pizza2004:
I noticed that when you place a slab in the bottom half of a water source block it wouldn't stop the downward flow of water.  Say I had a single block large hole with water flowing down out of it, I would expect the slab to make all the water go away, but instead nothing would change.  Turns out that's because it wasn't checking the down facing on the block above it when determining if it should disappear, only the up facing on itself.


tyronx:
There's bugs I would have never expected to exist, lol. This seems to only partially fix the issue. Instead of completely ignoring the slab it now stops the water. It should probably spread sideways. Merged either way, thanks!

pizza2004:
Oh strange.  I never tested it without blocks on the sides so I didn’t even realize it would stop it from flowing off the sides.  I’ll look into it!

PR #84, opened 2024-11-20
Fix bucket variants having broken model and textures

Craluminum2413:
When adding new bucket, it doesn't use new model and textures, but uses vanilla bucket model and textures instead.

This fix caches every variant separately, instead using single cache key for many variants, thus allowing custom buckets to display correctly


tyronx:
merged 👍

PR #85, opened 2024-11-20
Allow Jam Naming to be More Dynamic

pizza2004:
This change should allow for specific custom names for mixed jams as well as to change the default "automatic" name for a specific fruit when added to jam, so that it is possible to use "pineapple-in-jam-name" in the language file and make both "Pineapple Jam" and "Pineapple-cranberry jam" work instead of having mixed jams become something like "Pineapple slices-cranberry jam" instead.


tyronx:
thanks! merged.

PR #86, opened 2024-11-23
Use Majority Material for Microblocks Instead of Original Material Whenever Possible

pizza2004:
This change should make it so that you cannot do things like chisel a stone and combine it with glass in order to create a fully insulating block of glass.


tyronx:
hm. This would require caching of the majority material index in a single byte field, calculcated everytime the microblock changes. Otherwise this would have a notable performance impact.

pizza2004:
If it causes a performance issue it's definitely not worth it.  It's an obscure exploit anyway.

PR #87, opened 2024-11-23
Add Necessary Code for Labeled Trunk

pizza2004:
I made a mod that adds the Labeled Trunk model that has been seen in Dev Logs as an actual working [Labeled Trunk](https://mods.vintagestory.at/labeledtrunk) block and I figured I'd submit a PR that makes the necessary changes for that to work.

I also included bits that were used to make the Labeled Containers mod properly offset the text for the sign, in the form of the labelOffset variable, which would allow any GenericTypedContainer to be given the capability of having a sign added to it if those values could be set with the JSON.  For this PR it's just hard coded to what it would need to be for the Labeled Trunk model that's already in the game files.

If the trunk JSON file is copied and the block code is changed to "labeledtrunk", and all the classes, shapes, and textures updated properly it will simply work with this code added, nothing else needs to be changed, but you could allow labelOffset to be set by the JSON itself so it's more versatile, and you don't have to use a specific block code.

### DIFF
--- a/Block/BlockLabeledChest.cs
+++ b/Block/BlockLabeledChest.cs
@@ -46,4 +46,46 @@ namespace Vintagestory.GameContent
             return interactions.Append(base.GetPlacedBlockInteractionHelp(world, selection, forPlayer));
         }
     }
+
+    public class BlockLabeledTrunk : BlockGenericTypedContainerTrunk
+    {
+        WorldInteraction[] interactions;
+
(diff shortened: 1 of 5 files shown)


Craluminum2413:
Why not draft it?

pizza2004:
It's just not needed as code so for now it will just stay as a mod I suppose.

PR #91, opened 2024-12-14
Add distillation properties to handbook

jacopouggeri:
Currently distillation properties are not included in the handbook, this pr should address that.
Note:
- The text "One liter distills into" could be changed into simply "Distills into" and the stack size could be hidden. In the proposed change the ratio is used to determine how much one liter distills into, but this could be ignored if one so wishes.
- Currently brandy (spiritportion) is not present in the handbook since it has the handbook exclude attribute. This should be changed.

Extra fix: Adjusted juiceable itemstack slideshow, which used the bakeables array instead of the juiceables one


jacopouggeri:
Squashed version of https://github.com/anegostudios/vssurvivalmod/pull/78

PR #93, opened 2025-01-07
fix(Beehive): Fixed wrong temperature being used for hibernation

Rikaisan:
Fixed wrong temperature being used for hibernation (was using 5°C instead of 0°C as defined in the v1.15-pre.1 changelog) due to a wrong check on the activityLevel

The hibernation check is done at line 192: `if (actvitiyLevel < 1) return;` which was changed to `if (actvitiyLevel == 0) return;`

Also, the variable had a typo (should be `activityLevel` instead of `actvitiyLevel`)


pizza2004:
This has been merged.  Thank you.

PR #94, opened 2025-01-09
Use standard wildcard system to parse unsuitable codes

sekelsta:
Same change as https://github.com/anegostudios/vsessentialsmod/pull/13 but for troughs.


pizza2004:
This is purposefully like this for performance reasons, but thank you for the Pull Request!

PR #95, opened 2025-01-12
Give back stick after squeezing cheese

Lugrol:


Guimoute:
That is lovely!

Craluminum2413:
Stick is added to nutrition value

pizza2004:
Thank you, this will be included in a future release.

PR #96, opened 2025-01-16
Make private stuff protected in henbox

sekelsta:
This would mean classes inheriting from the henbox, like More Animals' nest box, can reuse more of the logic from their parent.


pizza2004:
Should be in a future release.

PR #97, opened 2025-01-20
Fix rotation bug in BlockEntity/BEAnvil.cs

TheMas3212:
Fixes anegostudios/VintageStory-Issues#4851


pizza2004:
Thanks for the code!  This was merged to the master branch and should be included in a future release.

PR #98, opened 2025-01-22
Use regex to search the handbook for all words separately

uramer:
See https://discord.com/channels/302152934249070593/351624415039193098/1331565708265390132 for the original discussion

This makes the search results much more intuitive. For example, "plate iron armor" actually shows the relevant armor.

One downside is that even very weak results (e. g. a single match in the page's body)  are added to the list. It doesn't seem to cause significant performance issues, but might be a bit confusing. so I'm welcome to any suggestions of limiting that (e. g. at most 50 matches, but always include all results that have matches in the title).

I've tested international support only in Russian, as that was easily accessible to me, but I don't think there is any functional difference from the previous implementation there.


Systems/Handbook/Tutorial/GuiHandbookTutorialPage.cs
```diff
@@ -149,16 +149,6 @@ public override void ComposePage(GuiComposer detailViewGui, ElementBounds textBo
...
-        public override float GetTextMatchWeight(string text)
-        {
-            /*string title = TextCacheTitle;
-            if (title.Equals(searchText, StringComparison.InvariantCultureIgnoreCase)) return 3;
-            if (title.StartsWith(searchText, StringComparison.InvariantCultureIgnoreCase)) return 2.5f;
-            if (title.CaseInsensitiveContains(searchText)) return 2;
-            if (TextCacheAll.CaseInsensitiveContains(searchText)) return 1;*/
-            return 0;
-        }
-
```
uramer, line 161: I did not add the same override as I was not sure it's intended purpose, but obviously it's not difficult to bring back.

pizza2004: `GetTextMatchWeight` is intended to return weighted results based on things like titles being higher in the list than descriptions.  Does your code adequately replace this weighting to make sure different things have different weights to them in the results?

uramer: I understand the purpose of `GetTextMatchWeight`, I was specifically referring to the commented out code.

Yes, it sorts all pages by title matches first, then breaks ties by page content matches, and then by title length to prefer more exact matches.

PR #99, opened 2025-01-23
Fill bloomery 1 ingot worth of nuggets at a time when sneaking

Guimoute:
Old behaviour: add 1 item normally, 5 when sneaking.

New behaviour: add 1 item normally, 20 when sneaking and it's nuggets, 5 when sneaking and adding other items (e.g. quartz for glass).

### DIFF
--- a/Block/BlockBloomery.cs
+++ b/Block/BlockBloomery.cs
@@ -150,7 +150,8 @@ namespace Vintagestory.GameContent
             {
                 
                 if (hotbarstack == null) return true;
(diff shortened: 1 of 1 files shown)


pizza2004:
Could you please change this so it is an attribute that can be specified in a JSON and provide the relevant change to the relevant JSON here as a code comment as well?

Guimoute:
I do not know how to convert hardcoded values into JSON-driver values. That's something I would like to learn.

PR #100, opened 2025-01-25
Improve fruit press

glutzer:
Can now add fractional amounts of wet mash.
Fixed bug where it randomly stops working until you break it.
Animations are client only now.
Removed global key listener.
Now prevents actions until fully unscrewed.

### DIFF
--- a/Block/BlockLantern.cs
+++ b/Block/BlockLantern.cs
@@ -322,8 +322,8 @@ namespace Vintagestory.GameContent
         {
             if (!byPlayer.Entity.Controls.ShiftKey)
             {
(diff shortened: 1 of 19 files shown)


pizza2004:
Luckily the bug where it randomly stops working is already fixed and will be included in the next version, so we don't need to wait on reviewing this for that to be fixed.

Can you comment on why you removed the stuff like proportional speed for juicing, extra juice dripping out when the screw is released, and `juiceableLitresTransfered` being added to the machine when inserting a mash?  I know in the case of the last it will cause incorrect players to lose dry mash if they add a partial mash, if not programmed correctly, plus you missed updating the variable you pass to the audit so it would just say "0x" there.

PR #102, opened 2025-02-21
Fix dry mash output randomness when pressedDryRatio != 0

Spoonail-Iroiro:
Fix  anegostudios/VintageStory-Issues#5360


pizza2004:
Thanks!  This will be in 1.20.5.

PR #103, opened 2025-02-22
Fix the crash on merging custom mashes without `transitionableProps` in a fruitpress

Spoonail-Iroiro:
Fix anegostudios/VintageStory-Issues#5366


pizza2004:
Thanks!  This will be in 1.20.5.

PR #104, opened 2025-02-24
Scale surface rock chance by deposit depth to world sea level.

mrgrim:
Resolves issue  https://github.com/anegostudios/VintageStory-Issues/issues/5378


pizza2004:
Thank you!  This will be in a future version.

PR #105, opened 2025-02-24
Initialize world sea level earlier for deposit generator dependency.

mrgrim:
Resolves issue https://github.com/anegostudios/VintageStory-Issues/issues/5386


pizza2004:
Thanks for the PR!  We have merged this internally and it will be in a future version.

PR #106, opened 2025-04-01
Allow honey comb squeezing into a barrel

IKnewOne:
Would need to alter `game:itemdesc-honeycomb`


pizza2004:
This was merged and will be available in the future.

PR #107, opened 2025-04-09
Synchronize deleted items for randomizers

wispae:
Fixes an issue wherein deleting items from the randomizer pool is not properly synchronized back to the server for the following:

- BlockEntityBlockRandomizer
- ItemLootRandomizer


pizza2004:
Could you please explain the issue this solves and why it needs to be solved?  Is it an issue in vanilla or only with mods?

pizza2004:
Ahh, this is a bug with using the interface to mess with it in creative and such rather than when it's being loaded?  Okay, that's a good fix then, thank you. :)

pizza2004:
Thanks for this, it will be in a future release!

PR #108, opened 2025-04-14
Allow IAnvilWorkable on CollectibleBehaviors (using GetCollectibleInterface)

jacopouggeri:
This change replaces direct casting of collectibles to `IAnvilWorkable` with calls to `GetCollectibleInterface<IAnvilWorkable>()`. This increases flexibility by allowing implementations of IAnvilWorkable via collectible behaviors instead of being limited to direct class implementations. It should also be compatible with any existing code.


pizza2004:
Thanks, this got merged into master and will be available in the future. :)

PR #110, opened 2025-04-22
Fix Monsters spawning in daylight

jabermony:
This fixes monsters spawning in daylight in
games where temporalStability is set to false.
This was achieved by modifying the existing logic
to perform the same operations in both cases but
assume a "temporal modifier" of 1 `mod = 1` when
temporalStability is false and
spawnCloserDuringLowStability is true.

For further detail on the issue see:
https://github.com/anegostudios/VintageStory-Issues/issues/5778
https://discord.com/channels/302152934249070593/1363629052723986603


tyronx:
thanks for the attempt, I this change should fix it as well, will be in the next release

image

PR #111, opened 2025-04-30
Call `CollectibleBehavior`s' corresponding methods in `ItemBow`'s `OnHeldInteractStop` and `OnHeldInteractCancel`

JohnDowson:
Many of the tools mostly ignore the `CollectibleBehavior`s they have.
Here's why that's a problem:

1. This makes `CollectibleBehavior`s useless, since none of the callbacks will be called with any sort of consistency
2. Debugging is painful, since each item has special handling. You can spend hours debugging new code, only to discover that one specific item just chose to violate the contract of calling it's Behaviour
4. Hinders modding, by making it hard to implement sweeping changes that touch multiple items.
5. Hinders mod compatibility, since the modders will employ fragile work-arounds to implement their desired changes anyway
7. The documentation is quite literally lying to both modders and game developers.
    This is what `CollectibleBehavior.OnHeldAttackStep`'s documentation says:
    `/// Called continously when a custom attack action is playing. Return false to stop the action.`
    This is patently not true, since many items including shovels, axes, and pickaxes simply never invoke this method in their `CollectibleObject.OnHeldAttackStep`

### DIFF
--- a/Item/ItemBow.cs
+++ b/Item/ItemBow.cs
@@ -121,6 +121,23 @@ namespace Vintagestory.GameContent
 
         public override bool OnHeldInteractCancel(float secondsUsed, ItemSlot slot, EntityAgent byEntity, BlockSelection blockSel, EntitySelection entitySel, EnumItemUseCancelReason cancelReason)
         {
+	        bool result = true;
+			bool preventDefault = false;
+			foreach (CollectibleBehavior collectibleBehavior in this.CollectibleBehaviors)
+			{
+				EnumHandling handled = EnumHandling.PassThrough;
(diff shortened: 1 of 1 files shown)


pizza2004:
Why would we not just call the base method instead of this?

JohnDowson:
The base method for `OnHeldInteractStop` has some funky behaviour that is too bizarre for me to understand:
```cs
			this.tryEatStop(secondsUsed, slot, byEntity);
```

It probably does makes sense to call the base method for `OnHeldInteractCancel`, I'll give you that.

sekelsta:
For these tools `tryEatStop` will get null from `GetNutritionProperties` and return without doing anything. Skipping it just prevents anyone from modding in edible pickaxes or whatever.

PR #114, opened 2025-06-15
Add EntityBehaviorGait, update EntityBehaviorRideable to utilize it

safwyls:
- Adds an extensible gait behavior for entities, allowing multiple movement paces (e.g. walk, trot, canter, gallop)
- GaitMeta object properties include
    - Turn radius
    - Move speed
    - Backwards flag
    - Fallback gait (gait item for transitioning to a fallback with certain actions)
    - Sprint flag
    - Sound asset location
- Updates rideable behavior to utilize these gaits allowing transitioning through available gaits as defined in rideable behavior config
- Supports any number/name of gaits (i.e. doesn't need to be named walk, trot etc. and doesn't need to be four gaits can be 1 to n gaits)
- Rideable utilizes GaitMeta movement speed with a multiplier such that an entity with a rider can scale the movement speed appropriately

See elk-tamed_withgait.json and elk-semitamed_withgait.json files for example behavior configs for EntityBehaviorGait and EntityBehaviorRideable

[elk-semitamed_withgait.json](https://github.com/user-attachments/files/20804222/elk-semitamed_withgait.json)
[elk-tamed_withgait.json](https://github.com/user-attachments/files/20804223/elk-tamed_withgait.json)

### DIFF
--- a/Entity/Behavior/BehaviorGait.cs
+++ b/Entity/Behavior/BehaviorGait.cs
@@ -1,0 +1,83 @@
+﻿using System.Linq;
+using Vintagestory.API.Client;
+using Vintagestory.API.Common;
+using Vintagestory.API.Common.Entities;
+using Vintagestory.API.Datastructures;
+
+namespace Vintagestory.GameContent
+{
+    public record GaitMeta
+    {
+        public string Code { get; set; } // Unique identifier for the gait, ideally matched with rideable controls
+        public float TurnRadius { get; set; } = 3.5f;
+        public float MoveSpeed { get; set; } = 0f;
+        public bool Backwards { get; set; } = false;
+        public float StaminaCost { get; set; } = 0f;
+        public string FallbackGaitCode { get; set; } // Gait to slow down to such as when fatiguing
+        public bool IsSprint { get; set; } // Used to toggle entity.Controls.Sprint from rideable, consider alternatives?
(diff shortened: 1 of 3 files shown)


safwyls:
found a semantic issue with tracking swimming gaits, going to move to draft until i work that out. hopefully later today or tomorrow

safwyls:
Recent Changes:

- Properly dispose of gait sounds so we dont fill the OpenAL buffer
- Use sprint control instead of ctrlkey for adjusting gait
- If only two gaits respect toggle sprint such that releasing sprint slows to previous gait (note: sprint gait in json must have IsSprint: true flag set for this to work)
- Fix bug with backing out of water triggering walk instead of walkback gait
- Change turn radius in gait configurations to yaw multiplier (more accurate name)
- Apply yaw multiplier to mount turning calculations such that it actually impact riding movement
 

Updated the elk entity json to match existing rideable gaits and use new yawMultiplier config in gaits. The gaits include all possible movements that elk has in vanilla, but rideable is currently restricted to just walk/sprint to match the existing setup. This could be changed later if desired.

[elk-semitamed.json](https://github.com/user-attachments/files/21153813/elk-semitamed.json)
[elk-tamed.json](https://github.com/user-attachments/files/21153814/elk-tamed.json)

PR #115, opened 2025-06-23
Use trough suitability in BETrough.GetBlockInfo

Lazula:
Fixes anegostudios/VintageStory-Issues#6010 and anegostudios/VintageStory-Issues#4462

I also got rid of the "creatureDietGroup" check since it looks like a remnant of old code and that name doesn't appear anywhere else.

### DIFF
--- a/BlockEntity/BETrough.cs
+++ b/BlockEntity/BETrough.cs
@@ -392,41 +392,40 @@ namespace Vintagestory.GameContent
             dsc.AppendLine(Lang.Get("Portions: {0}", fillLevel));
 
             ItemStack contentsStack = config.Content.ResolvedItemstack ?? ResolveWildcardContent(config, forPlayer.Entity.World);
-            
-            if (contentsStack != null)
+
+            if (contentsStack == null) return;
+
+            var cobj = contentsStack.Collectible;
+            var foodCat = cobj.NutritionProps?.FoodCategory ?? EnumFoodCategory.NoNutrition;
+            var foodTags = cobj.Attributes?["foodTags"].AsArray<string>();
+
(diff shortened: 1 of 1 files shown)


BlockEntity/BETrough.cs
```diff
@@ -392,41 +392,40 @@ public override void GetBlockInfo(IPlayer forPlayer, StringBuilder dsc)
...
 
-                HashSet<string> creatureNames = new HashSet<string>();
-                foreach (var entityType in Api.World.EntityTypes)
+            dsc.AppendLine(Lang.Get(contentsStack.GetName()));
+
+            HashSet<string> creatureNames = new HashSet<string>();
+            foreach (var entityType in Api.World.EntityTypes)
+            {
+                var attr = entityType.Attributes;
+                if (attr == null || attr["creatureDiet"].Exists == false) continue;
```
pizza2004, line 408: Just FYI this could simply have been written as `if (attr?["creatureDiet"].Exists != true) continue;` instead to not need the null checking separated. :)

pizza2004:
Thanks!  I’ll look at this in a little bit. :)

BlockEntity/BETrough.cs
Lazula: Thanks, new language + framework means I don't have the most concise terms yet!

pizza2004: Oh no worries I wasn’t really asking you to change it anyway.  I always have to clean the code up from PRs anyway when I merge them in I was more just making you aware for your own sake.  Can you place squash the commits?  It’s a pain to work with PRs that have more than 1 commit. :)

pizza2004:
This will be included in the next prerelease.

PR #116, opened 2025-06-28
Refactor henbox for generalized item storage and display, misc bugfixes

sekelsta:
BEHenBox now inherits from BEContainerDisplay, allowing it to store and display arbitrary items using the same code that shelves and display cases do.
 - Arbitrary items can be rendered and displayed. For instance if the game added quail and they laid speckled eggs, those could be displayed in any combination with the white chicken eggs.
 - The BlockEntity includes an inventory, rather than a unique block type for each combination of items.
   - A container with constant perish rate is used, both because it runs faster and because if egg spoilage were added a constant rate would be more realistic. Broody hens would be keeping the eggs at a relatively constant temperature regardless of the environment, plus raising chickens in cellars would be in poor taste.
 - Taking eggs from the hen box now goes through Block.OnBlockInteractStart instead of BlockBehavior.OnBlockInteractStart so that it happens both serverside and clientside, keeping inventories correctly synced.
 - Data about the chick (entity code, generation, etc) is now stored in the egg's item attributes, completing a TODO in the code and making progress towards allowing the player to move eggs between nestboxes. Note that actually allowing players to place eggs in nestboxes would be OP unless the perish rate is set to something non-zero, a design change out of scope of this particular PR.
   - Chick data includes herd ID, which should simplify the merge with Maltiez's changes
   - Eggs ignore and erase chick data when merging itemstacks, making the change invisible to players
 - Note BlockBehavior CollectFrom is now unused, but is left in on advice of pizza2004
 - Both fertile and infertile eggs transition over correctly from previous save format to this one
   - Henboxes held in an inventory will stay in the prior format until placed. If this is undesired, I can adjust this to keep "henbox-empty" as the henbox code instead, though it would be an inaccurate description.

Bugfixes:
 - Block info for a full hen box with fertilized eggs always showing "A broody hen is needed" even when a broody hen was present. The value was checked clientside but only set serverside. Fixes at least part of https://github.com/anegostudios/VintageStory-Issues/issues/5307
 - If the hen box has a mix of fertile and infertile eggs, the infertile eggs will no longer be destroyed when the fertile ones hatch
 - [Modding API] Use AssetLocation.Create instead of entity.Code.CopyWithPath, allowing entities to lay eggs that hatch into something from a different mod domain.

Asset changes:
 - `assets/survival/entities/land/chicken-adult.json`: Added line to attributes ` eggTypes: [ { type: "item", code: "game:egg-chicken-raw" } ],`
 - `assets/survival/recipes/grid/henbox.json`: Recipe output changed to "henbox"
 - `assets/survival/blocktypes/wood/henbox.json` and `assets/survival/blocktypes/legacy/henbox.json`: Major changes, see included files
 - Shape files survival/shapes/block/wood/henbox/henbox-(1egg/2eggs/3eggs).json can be deleted
 - Lang needs to be updated, not included.

### DIFF
--- a/Block/BlockHenbox.cs
+++ b/Block/BlockHenbox.cs
@@ -7,9 +7,19 @@ namespace Vintagestory.GameContent
 {
     public class BlockHenbox : Block
     {
+        public override bool OnBlockInteractStart(IWorldAccessor world, IPlayer byPlayer, BlockSelection blockSel) {
+            var blockEntity = world.BlockAccessor.GetBlockEntity(blockSel.Position) as BlockEntityHenBox;
+            if (blockEntity != null) {
+                return blockEntity.OnInteract(world, byPlayer, blockSel);
+            }
+
+            return base.OnBlockInteractStart(world, byPlayer, blockSel);
+        }
+
         public override WorldInteraction[] GetPlacedBlockInteractionHelp(IWorldAccessor world, BlockSelection selection, IPlayer forPlayer)
         {
-            if (Variant["eggCount"] == "empty") return System.Array.Empty<WorldInteraction>();
+            var blockEntity = world.BlockAccessor.GetBlockEntity(selection.Position) as BlockEntityHenBox;
+            if (blockEntity == null || blockEntity.CountEggs() == 0) return System.Array.Empty<WorldInteraction>();
 
             return new WorldInteraction[]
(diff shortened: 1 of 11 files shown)


pizza2004:
This will be in the next prerelease.

PR #117, opened 2025-06-28
Fix issues with containers that have variantGroups

Craluminum2413:
For 1.21+ only!

Sometimes when storage vessel falls, you can briefly notice that it has different textures.
It happens because only FirstCodePart is cached instead Code


pizza2004:
Fixed this another way but thanks for pointing it out!

PR #118, opened 2025-07-02
Add nest type for determining which birds can use which nests (VSSurvivalMod part)

sekelsta:
Together with a PR to VSEssentials (https://github.com/anegostudios/vsessentialsmod/pull/24), this does:

Modding API Feature: Species other than chickens are able to use the hen box if they use a matching nest type
Modding API Feature: AI task seekblockandlay now supports multiple chick types (such as male and female)
Modding API Tweak: IAnimalNest.TryAddEgg now takes an egg itemstack with chick info set in attributes

I'm including the previous merged PR underneath in the hopes that git will apply it correctly without extra merge conflicts. If that works correctly, the actual changes should look like this: https://github.com/anegostudios/vsessentialsmod/compare/master...sekelsta:vsessentialsmod:henboxPR?w=1 (plus potentially some line ending changes)

### DIFF
--- a/Block/BlockHenbox.cs
+++ b/Block/BlockHenbox.cs
@@ -7,6 +7,17 @@ namespace Vintagestory.GameContent
 {
     public class BlockHenbox : Block
     {
+        public string NestType;
+
+        public override void OnLoaded(ICoreAPI api)
+        {
+            base.OnLoaded(api);
+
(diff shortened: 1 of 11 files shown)


pizza2004:
Will be in pre.2!  Thanks!

PR #120, opened 2025-07-08
Move PitKiln base.Initialize() before the BEBurning callback registration

Fayti1703:
This fixes https://github.com/anegostudios/VintageStory-Issues/issues/5181.

Because `BlockEntityGroundStorage` registers its own `OnFireDeath` callback as of 1.20.0-pre.1 (cf0900e82304b5c42128ef7259ee5dd1e49a3b44), it overrides the behavior in PitKiln that resets the "lit" property, causing the issue mentioned above.

I'm not sure if changing this can cause other issues, but a quick test didn't show any odd behavior after applying the patch.


pizza2004:
Thank you for this!  I couldn't ever figure out why that bug was happening.  This will be in the next release.

PR #125, opened 2025-07-17
Extract Shield Logic to ItemShieldBase for Better Moddability

jacopouggeri:
This PR introduces a new `ItemShieldBase` class, extracting all core shield logic to make shield items more moddable and flexible. This also removes hardcoded limitations that previously limited shields to only the vanilla item type, limiting content modders. With this change, shields such as the crude and blackguard variants could be migrated to use this base class, and content modders can now create custom shields more easily. 

Key changes:

- Extracted core shield logic into a new base class, `ItemShieldBase.cs`
- Cleaned up `ModSystemStopRaiseShieldAnim`
- Replaced loose strings with constants where appropriate
- Added an `isAttachable` attribute to shield JSON, allowing shields to be defined as attachable or not (default: `true`)

This refactor should help both core development and the modding community by making shield implementation more modular and maintainable.

### DIFF
--- a/Item/ItemShield.cs
+++ b/Item/ItemShield.cs
@@ -15,64 +15,28 @@ using Vintagestory.API.Util;
 
 namespace Vintagestory.GameContent
 {
-    public class ModSystemStopRaiseShieldAnim : ModSystem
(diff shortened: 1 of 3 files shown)


pizza2004:
This will be in a future version.

PR #126, opened 2025-07-18
List bookshelves, scroll racks, antler mounts in handbook

ATP-Kraken:
Items compatible with the aforementioned storage blocks do not say so in the handbook.
This change adds handbook links for these storage methods, improving consistency.
For example, papers and tuning cylinders aren't scrolls, but can still be stored in a scroll rack.

Example localization for these keys:
```
	"handbook-storable-bookshelf":  "• <a href=\"handbooksearch://bookshelf\">Bookshelves</a>",
	"handbook-storable-antlermount": "• <a href=\"handbooksearch://antler mount\">Antler mounts</a>",
	"handbook-storable-scrollrack": "• <a href=\"handbooksearch://scroll rack\">Scroll racks</a>",
```
I suggest a search rather than a direct link because these blocks come in multiple variants.


pizza2004:
Thanks!  This will be in the next release.

PR #127, opened 2025-07-18
Allow Containers and Contents to set Custom Drink Speed

jacopouggeri:
This PR allows to make containers and their contents more moddable by allowing custom drink speed and modifiers.
Containers can now define a `drinkSpeed` attribute (defaults: `1`) which sets the liquid quantity consumed when drinking. This can be defined just like `capacityLitres` either as a general attribute or as a part of `liquidContainerProps`. (This solution was chosen for consistency with the current code, alternatively one could adopt the second option only to have less ambiguity). `WaterTightContainableProps` have been also updated to allow a drink speed modifier, which would allow certain liquids (perhaps honey?) to have lower drink speed. At the moment this value used by multiplying the current drink speed, so that the liquid itself doesn't override the container's speed, and both are taken into account.

Arguably, drink speed should be capped to 1, but that could be achieved by always choosing to use values lower than one for both attributes. This is true for vanilla at the moment (obviously), and modders could potentially benefit from the freedom of a higher value in niche situations, so I haven't capped it here.

To avoid liquids being impossible to drink at all if the minimum amount of litres is higher than the drink speed, the actual drink speed value is capped from below by `1.0f/ItemsPerLitre`.

### DIFF
--- a/Systems/Liquid/BlockLiquidContainerBase.cs
+++ b/Systems/Liquid/BlockLiquidContainerBase.cs
@@ -17,7 +17,9 @@ namespace Vintagestory.GameContent
     public abstract class BlockLiquidContainerBase : BlockContainer, ILiquidSource, ILiquidSink
     {
         protected float capacityLitresFromAttributes = 10;
+        protected float drinkSpeedFromAttributes = 1;
         public virtual float CapacityLitres => capacityLitresFromAttributes;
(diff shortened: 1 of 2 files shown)


pizza2004:
A slightly modified version of this has been merged and will be in a later version of the game.

PR #129, opened 2025-07-19
Fix curb bit behavior in pre.3

safwyls:
Fixes https://github.com/anegostudios/VintageStory-Issues/issues/6301


pizza2004:
Included in the next release.

PR #130, opened 2025-07-21
Fix a null or empty check in BlockMeal.cs

traugdor:
fixed a null or empty check that was just really bothering me


pizza2004:
This change was made to the code, now.

PR #137, opened 2025-09-23
Fix BlockEntityBookshelf crash

jayu-vs:
Fixes [this crash](https://pastebin.com/7YDP3Nvh) due to null book shelf Type.
Discord mention: https://discord.com/channels/302152934249070593/402173001178677268/1419125595077345321

### DIFF
--- a/BlockEntity/BEBookshelf.cs
+++ b/BlockEntity/BEBookshelf.cs
@@ -33,10 +33,9 @@ namespace Vintagestory.GameContent
 
         public int[]? UsableSlots {
             get {
-                var bs = block as BlockBookshelf;
-                if (bs == null) return System.Array.Empty<int>();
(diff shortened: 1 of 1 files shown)


BlockEntity/BEBookshelf.cs
```diff
@@ -33,10 +33,9 @@ public float MeshAngleRad
 
         public int[]? UsableSlots {
             get {
-                var bs = block as BlockBookshelf;
-                if (bs == null) return System.Array.Empty<int>();
-                bs.UsableSlots.TryGetValue(Type!, out var slots);
-                return slots;
+                if (block is not BlockBookshelf bs || Type == null) return System.Array.Empty<int>();
```
Ilan321, line 36: `Array.Empty()` can be replaced with `[]`

jayu-vs:
Looks like this was fixed in 1.21.2

PR #138, opened 2025-10-12
Update MechanicalPowerMod, BEClayOven

tehtelev:
**Issue № 1.** This patch introduces a fix for the null error after CreateJoinAndDiscoverNetwork. This is due to the fact that some of my devices connect to the mechanical network after the initialization phase. Therefore, sometimes, especially at the boundaries of loaded chunks, CreateJoinAndDiscoverNetwork can return null . Adding this protection doesn't change anything. However, my devices from the mod work perfectly and don't interfere with chunk loading.
[Experiencing the old bug, probably the same as in Electricity](https://github.com/tehtelev/ElectricalProgressive/issues/5)


sekelsta:
The MechanicalPowerMod change went in separately in 1.22.0-rc.9

PR #173, opened 2026-01-27
ItemRoller rotation is inconsistent

Darthsack:
Found this bug when working on the airship mod
When facing in the default direction it displayed and place offset by one
Though they are consistent with each other
They don't line up on the placed block on the model

Adding a few new states to the  placement code as well the model rotator fixes the bug in code, though a json change will be needed to realine the model


pizza2004:
Could you show exactly what it is this solves?  I'm struggling to understand your explanation.

pizza2004:
Could you also show me the necessary JSON change for this?

pizza2004:
This should be implemented for 1.22.

PR #174, opened 2026-01-31
Average transition states of combined liquids in BlockLiquidContainerBase.TryPutLiquid and TryMergeStacks

### LINKED ISSUE #2839: Liquid freshness of the whole barrel is defined by the first liquid poured instead of an average
### Game Version

v1.18.5

### Platform

Windows

### Modded

Modded

### SP/MP

Singleplayer

### Description

(None of my mods affect barrels)

As shown in the video, if you pour any amount of fresh liquid in a barrel, the spoiled liquid subsequently added is considered fresh and adopts the remaining shelf life of the initial liquid. Althought untested, I assume it works the other way around too.  

https://github.com/anegostudios/VintageStory-Issues/assets/51537797/60fb1718-c6b0-4cbf-a4c4-31980747a86a

### How to reproduce

_No response_

### Screenshots

_No response_

### Logs

```Log```

Lazula:
Fix anegostudios/VintageStory-Issues#2839 and anegostudios/VintageStory-Issues#1569 by averaging perish states.

Vanilla behavior:

https://github.com/user-attachments/assets/02181c41-d365-478e-bc33-cbe59847846d

Fixed behavior:

https://github.com/user-attachments/assets/87590e5c-ffd7-4f5b-a701-3df7ca9d22bb

### DIFF
--- a/Systems/Liquid/BlockLiquidContainerBase.cs
+++ b/Systems/Liquid/BlockLiquidContainerBase.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Text;
 using Vintagestory.API.Client;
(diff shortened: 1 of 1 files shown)


Lazula:
A quick demo of the averaging working correctly since I only combined equal amounts in the video

 `((10 * 24) + (20 * 96)) / (10 + 20) = 72`

https://github.com/user-attachments/assets/572c991c-1836-4e58-86ed-14c5453cfdad

Lazula:
The divide by zero check is necessary in case something tries to move 0 items of liquid into an empty container which would produce a NaN freshness value and a crash on consumption.

Currently happens in the fruitpress if you stop juicing at just the right time to make the mash show up as dry due to having less than 0.01L left. You can still try to press out the fraction of a liquid item, but since it can't round up it will never succeed and keep trying to move 0 items, which is is its own issue that should be patched separately.

Lazula:
Now covers all transition states and TryMergeStacks for combining in inventory.

Lazula:
Included in 1.22.0-rc.4

PR #176, opened 2026-02-04
Prevent taking one portion of liquid being overwritten by drinking

### LINKED ISSUE #5840: Trying to take more than one portion of consumable liquid is overridden by eating action
### Game Version

v1.20.9

### Platform

Windows

### Modded

Modded

### SP/MP

Singleplayer

### Description

When you attempt to take more than one portion by shift-clicking with any container (e.g. bucket, jug) on another container that has consumable liquid (e.g. honey, juice, alcohol), your character will prioritize eating the portion you just got instead of taking more portions.

### How to reproduce

1. Get some container with consumable liquid: let's say a bucket with 10L of honey
2. Get an empty container to scoop the honey with: let's say a jug
3. Shift-click the bucket with the empty jug: 0.3L will be taken from the bucket, since they have a max capacity of 3L; if you tried this with another bucket, you'd be taking 1L
4. Try that again: you won't be able to take more, instead your character will attempt eating the 0.3L of honey

### Screenshots

_No response_

### Logs

```Crash```

Lazula:
Fix anegostudios/VintageStory-Issues#5840 and anegostudios/VintageStory-Issues#1697

Calling up to Collectible.OnHeldInteractStart too early was hijacking the control flow and branching into tryEatBegin, so we have to delay that until confirming that we don't need to do any liquid interactions. Shift+RMB when looking at the side of the block still results in ground storage.

It seems like liquid containers should maybe have block interaction priority, but I didn't want to mess with that and potentially create much bigger problems.

Vanilla:

https://github.com/user-attachments/assets/e73ec58d-852b-425f-a0fe-acaa7e9280bd

Fixed:

https://github.com/user-attachments/assets/31217836-301a-42c3-80ab-ec4807e125be

### DIFF
--- a/Systems/Liquid/BlockLiquidContainerBase.cs
+++ b/Systems/Liquid/BlockLiquidContainerBase.cs
@@ -686,15 +686,30 @@ namespace Vintagestory.GameContent
 
             if (blockSel == null || byEntity.Controls.ShiftKey)
             {
-                if (byEntity.Controls.ShiftKey) base.OnHeldInteractStart(itemslot, byEntity, blockSel, entitySel, firstEvent, ref handHandling);
+                // We need to make sure we aren't doing anything
+                // with liquid handling before calling up to
+                // the base method because it will call out to
+                // tryEatBegin and hijack our control flow.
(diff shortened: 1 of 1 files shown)


Lazula:
Included in 1.22.0-rc.4

PR #177, opened 2026-02-04
Always return pulverizer input item if caps are too weak

### LINKED ISSUE #3073: Pulverizer with insufficient tier pounder caps still destroys nuggets
### Game Version

v1.18.10

### Platform

Windows

### Modded

Vanilla

### SP/MP

Singleplayer

### Description

A pulverizer with iron pounder caps will destroy (2/3rds of) ilmenite nuggets, despite being unable to crush the remainder into powder (returning nuggets instead.) One would expect all nuggets to be returned if they cannot be crushed by the caps.

### How to reproduce

1. Assemble a pulverizer with iron pounder caps and power it.
2. Place ilmenite nuggets in the pulverizer as it operates.
3. Observe that the pulverizer usually returns nothing instead of a nugget.
4. Replace iron pounder caps with steel pounder caps.
5. Place ilmenite nuggets in the pulverizer as it operates.
6. Observe that the pulverizer usually returns nothing but creates powdered ilmenite as expected by the 3:1 conversion ratio.

### Screenshots

_No response_

### Logs

_No response_

Lazula:
Fix anegostudios/VintageStory-Issues#3073

Removed an early return in Crush() that caused there to never to be any item spawned if the randomized output stack was empty.

Vanilla:

https://github.com/user-attachments/assets/378ca44f-0112-4b35-9f55-f7e637c23af9

Fixed:

https://github.com/user-attachments/assets/2065c60f-42d0-4e2b-a639-269780431901

### DIFF
--- a/Systems/MechanicalPower/BlockEntity/BEPulverizer.cs
+++ b/Systems/MechanicalPower/BlockEntity/BEPulverizer.cs
@@ -138,30 +138,32 @@ namespace Vintagestory.GameContent.Mechanics
         private void Crush(int slot, int capTier, double xOffset)
         {
             ItemStack inputStack = inv[slot].TakeOut(1);
-            var props = inputStack.Collectible.GetCrushingProperties(Api.World, inputStack);
             ItemStack outputStack = null;
 
-            if (props != null) {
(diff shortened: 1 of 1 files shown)


Lazula:
Included in 1.22.0-rc.4

PR #181, opened 2026-03-01
Optimization of the firepit and the pot

tehtelev:
Improved cooking pot rendering in a campfire. The main issue is that a cooking pot is always rendered when it's in a campfire, even if nothing is being cooked in it. Another issue is that such campfires can also be found in ruins. Moreover, there may be dozens of such ruins in the rendering area, which the player may never find. However, the cooking pot will always render.
**Suggested improvements:**
1) The cooking pot renderer now has a clipping based on the player's view cone and the RenderRange distance. I chose a distance of 200 because at that distance, the campfire turns into just a few pixels on the screen.
2) Additionally, following Tyrone's suggestion, a system was implemented that renders a static mesh of the cooking pot when it's cold. When the temperature exceeds 50 degrees, the dynamic renderer will be activated for the cooking pot.

This makes it possible to reduce the impact of campfires on player tickrates to zero, including in settlements.

**My tests with 1000 blocks rendered**

Vanilla
1024 ванила

With revision
1024 мой

**Video of testing**
[Vanilla (test on 1.21.6)](https://youtu.be/6snnBBEkPyY)
[Visibility cutoff only](https://youtu.be/xEOmgS2TKlI)
[Visibility cutoff + static meshes](https://youtu.be/_7Bo6GG8qzo)

### DIFF
--- a/BlockEntity/Firepit/BEFirepit.cs
+++ b/BlockEntity/Firepit/BEFirepit.cs
@@ -18,7 +18,8 @@ namespace Vintagestory.GameContent
 
         // Temperature before the half second tick
         public float prevFurnaceTemperature = 20;
-
+        // flag to determine whether static or dynamic rendering is used
+        private bool prevForceStatic = false;
         // Current temperature of the furnace
         public float furnaceTemperature = 20;
         // Current temperature of the ore (Degree Celsius * deg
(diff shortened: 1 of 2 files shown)


tyronx:
Thanks a lot for this MR tehtelev. This has been merged into the 1.23 branch.

PR #183, opened 2026-03-09
Extract data manipulation code into separate methods for ore map UI creation

KnewOne:
The main intent behind this PR is to make ore map layer modifications easier by externalizing methods that provide data for ore map layer and its components into separate methods that mods can hook with harmony.
Additionally, sorts the dropdown values by user's locale.

Possible use cases are adding group-like tags ( iron-* which can be captured to allow display of limonite/magnetite/etc ), easier change of displayed text. 

Would simplify patching [dialog extras](https://github.com/KnewOne/BetterErProspecting/blob/main/BetterErProspecting/Patches/Stone/ComposeDialogExtrasPatch.cs), [ore map component](https://github.com/KnewOne/BetterErProspecting/blob/main/BetterErProspecting/Patches/Stone/OreMapComponentCtorPatch.cs) and [map rebuild](https://github.com/KnewOne/BetterErProspecting/blob/main/BetterErProspecting/Patches/Stone/RebuildMapComponentsPatch.cs) - transpilers would be fragile here since multiple mods would want to operate on this data batch

Also adds sorting of elements in the dropdown by their localized name. Respects "everyone" as the first option

### DIFF
--- a/Systems/Prospecting/OreMapLayer/OreMapComponent.cs
+++ b/Systems/Prospecting/OreMapLayer/OreMapComponent.cs
@@ -13,6 +13,7 @@ namespace Vintagestory.GameContent
     {
         Vec2f viewPos = new Vec2f();
         Vec4f color = new Vec4f();
+        PropickReading reading;
         int waypointIndex;
         Matrixf mvMat = new Matrixf();
         OreMapLayer oreLayer;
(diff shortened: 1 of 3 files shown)


Systems/Prospecting/OreMapLayer/OreMapComponent.cs
```diff
@@ -13,25 +13,37 @@ public class OreMapComponent : MapComponent
...
         Vec4f color = new Vec4f();
-        PropickReading reading;
         int waypointIndex;
         Matrixf mvMat = new Matrixf();
         OreMapLayer oreLayer;
         bool mouseOver;
 
         public static float IconScale = 0.85f;
         public string filterByOreCode;
+        public PropickReading reading;
```
pizza2004: A couple of things, for one, we try to capitalize our public fields (I know the one right above that isn't), and does this *need* to be publicly available to *set* and not just to *get*?

pizza2004 (changes requested):

tyronx:
thank you for the MR, this has been merged into the 1.23 branch

PR #188, opened 2026-03-30
Use CanEat for meals and liquid containers

Lazula:
Use the proposed CanEat method from anegostudios/vsapi#77 in meal and liquid container code.

Meals now use their default HeldTpUseAnimation if they are currently inedible.

It may be preferred for pies to have `twohandplaceblock` as their use animation now, since `interactstatic` looks a little bit weird.


pizza2004:
This was added for the future, thanks!

PR #190, opened 2026-04-04
Fixing mechanical block texture issues caused by texture atlases

tehtelev:
When installing a large number of mods, or when setting "maxTextureAtlasWidth" and "maxTextureAtlasHeight" to low values, the textures of mechanical blocks are displayed incorrectly. This is because the textures of the same model can be scattered across different texture atlases. By default, only one of them was always used.
What's been done:
- **Splitting the mesh by atlas.** SplitByTextureId() → one mesh turns into several, each with a single atlasTextureId.
- **Switching to multiple MeshRefs**. Instead of a single MeshRef, there's now a list—one for each atlas.
- **Correct texture binding.** Before each draw call, a specific atlasTextureId is bound, not a single shared one.
- **Preserved instanced rendering**. The same CustomFloats is used for all parts → animations and transformations are not broken.
- **Removed hard binding to atlas[0].** Rendering no longer depends on a single texture atlas → all textures are displayed correctly.

This solution may not be perfect, as I'm not very knowledgeable about graphics. But it works.

**Before:** 

**After**

### DIFF
--- a/Systems/MechanicalPower/Renderer/GenericMechBlockRenderer.cs
+++ b/Systems/MechanicalPower/Renderer/GenericMechBlockRenderer.cs
@@ -1,3 +1,5 @@
+using System;
+using System.Collections.Generic;
 using Vintagestory.API.Client;
 using Vintagestory.API.Common;
 using Vintagestory.API.MathTools;
@@ -8,10 +10,12 @@ namespace Vintagestory.GameContent.Mechanics
 {
     public class GenericMechBlockRenderer : MechBlockRenderer
     {
-        CustomMeshDataPartFloat matrixAndLightFloats;
-        MeshRef blockMeshRef;
+        private CustomMeshDataPartFloat matrixAndLightFloats;
+        private readonly List<MeshRef> blockMeshRefs = new List<MeshRef>();
+        private readonly List<int> blockMeshTextureIds = new List<int>();
 
-        public GenericMechBlockRenderer(ICoreClientAPI capi, MechanicalPowerMod mechanicalPowerMod, Block textureSoureBlock, CompositeShape shapeLoc) : base(capi, mechanicalPowerMod)
+        public GenericMechBlockRenderer(ICoreClientAPI capi, MechanicalPowerMod mechanicalPowerMod, Block textureSoureBlock, CompositeShape shapeLoc)
+            : base(capi, mechanicalPowerMod)
         {
             AssetLocation loc = shapeLoc.Base.Clone().WithPathPrefixOnce("shapes/").WithPathAppendixOnce(".json");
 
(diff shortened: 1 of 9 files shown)


Systems/MechanicalPower/Renderer/GenericMechBlockRenderer.cs
```diff
@@ -44,15 +48,38 @@ public GenericMechBlockRenderer(ICoreClientAPI capi, MechanicalPowerMod mechanic
...
 
-            this.blockMeshRef = capi.Render.UploadMesh(blockMesh);
+            MeshData[] splitMeshes = blockMesh.SplitByTextureId();
+
+            for (int i = 0; i < splitMeshes.Length; i++)
+            {
+                MeshData mesh = splitMeshes[i];
+                if (mesh == null || mesh.VerticesCount == 0) continue;
+
+                if (mesh.CustomFloats == null)
```
Kal-culated: This if else runs the same code in either path, can probably just remove the if else and just keep the mesh.CustomFloats set

Systems/MechanicalPower/Renderer/GenericMechBlockRenderer.cs
```diff
@@ -44,15 +48,38 @@ public GenericMechBlockRenderer(ICoreClientAPI capi, MechanicalPowerMod mechanic
...
+                {
+                    mesh.CustomFloats = matrixAndLightFloats;
+                }
+                else
+                {
+                    // keep the same instanced float layout for every split mesh
+                    mesh.CustomFloats = matrixAndLightFloats;
+                }
+
+                int atlasTextureId = (mesh.TextureIds != null && mesh.TextureIds.Length > 0)
```
Kal-culated: Can reduce lines like this with optional chaining, instead `(mesh.TextureIds?.Length > 0)`, same code just a little more compact

tyronx:
why not use `MultiTextureMeshRef`?

tyronx:
oh its not supported by the instanced renderer api

Systems/MechanicalPower/Renderer/GenericMechBlockRenderer.cs
```diff
@@ -66,20 +93,30 @@ public override void OnRenderFrame(float deltaTime, IShaderProgram prog)
...
+            updateMesh.CustomFloats = matrixAndLightFloats;
+
+            for (int i = 0; i < blockMeshRefs.Count; i++)
             {
-                matrixAndLightFloats.Count = quantityBlocks * 20;
-                updateMesh.CustomFloats = matrixAndLightFloats;
-                capi.Render.UpdateMesh(blockMeshRef, updateMesh);
-                capi.Render.RenderMeshInstanced(blockMeshRef, quantityBlocks);
+                prog.BindTexture2D("tex", blockMeshTextureIds[i], 0);
+                capi.Render.UpdateMesh(blockMeshRefs[i], updateMesh);
```
tyronx: updateMesh would also need to be split by textureId

tyronx: but also as mentioned in the comment, we ought to be using MultiTextureMeshRef  🤔

tyronx: wait nvm, i'm wrong.

tehtelev:
**Warning!** If you apply these changes, render types other than GenericMechBlockRenderer will have broken textures! Perhaps you can come up with a better solution for everyone!

tehtelev:
I've finalized my solution.
Each renderer splits its mesh(es) into groups by atlas if the textures are scattered across different atlases. Then, during rendering, it sends these meshes for rendering with the required textures. This logic is extracted into the MechBlockRenderer base class for these renderers to unify the code.

image

PR #194, opened 2026-05-03
Fix bugs in EntityBobber fishing logic

### LINKED ISSUE #9241: Fishing poles never decrease in durability
### Game Version

v1.22.0

### Platform

Mac

### Modded

Vanilla

### SP/MP

Singleplayer

### Description

Fishing poles have durability according to the tooltip. But the durability level never decreases - when cast, when fish are caught, with or without bait, or even when the line snaps due to distance. The same thing happens with both wood and bamboo poles.

### How to reproduce

1. Craft a new fishing pole.
2. Use it to catch fish, with or without bait, or snap the line due to distance.
3. Observe the durability remains at 300/300.

### Screenshots

_No response_

### Logs (Required)

[client-audit.log](https://github.com/user-attachments/files/27103980/client-audit.log)
[client-chat.log](https://github.com/user-attachments/files/27103984/client-chat.log)
[client-debug.log](https://github.com/user-attachments/files/27103982/client-debug.log)
[client-main.log](https://github.com/user-attachments/files/27103979/client-main.log)
[server-audit.log](https://github.com/user-attachments/files/27103978/server-audit.log)
[server-build.log](https://github.com/user-attachments/files/27103981/server-build.log)
[server-chat.log](https://github.com/user-attachments/files/27103983/server-chat.log)
[server-debug.log](https://github.com/user-attachments/files/27103986/server-debug.log)
[server-main.log](https://github.com/user-attachments/files/27103985/server-main.log)

averixus:
Hi, this is the first time I've tried contributing here so I'm not totally sure of the protocol. Please let me know if I need to do anything differently.

I started working on a mod that relates to fishing, and in the process I found a few bugs in the vanilla fishing code in `EntityBobber`. In fixing them, I also found some redundancies and inefficiencies in the logic and ended up substantially rewriting several methods to make them neater and clearer. 

I wrote these fixes originally as a harmony patch mod. I wasn't able to run the changed vssurvivalmod code directly, but I've tested the harmony version thoroughly, and verified as far as I can that it fixes the bugs and otherwise replicates vanilla behaviour. You can take a look at the harmony patch version [here](https://github.com/averixus/FishingConfig/blob/main/FishingConfig/Patches/EntityBobberPatches.cs).

#### Bugs fixed
These changes affect game behaviour.
- The bobber now correctly loses its catch and resets after 0.7s if the player doesn't reel in the line. The main error was that `catchAccum` never had the chance to increment properly, so the > 0.7s block would not execute. This meant that with the original code, the catch would be reeled in after any amount of time, even if it was an entity and had wandered away. 
- Fish entities will correctly flee the bobber as soon as the player misses the catch. Previously  `caughtFish` was being set to null before its value was checked, so the code inside would never be reached. The `"fleebobber"` task was also behind `ShouldExecute`, which seems to return false every time (possibly due to execution chance or priority?). This all meant that when a fish entity was released as a missed catch, it could immediately be caught again because it would stay where it was.
- The bobber now loses its bait if the player doesn't reel in the line in time. I'm not 100% sure this was the intended behaviour, but I _assume_ there was meant to be a penalty for missing a catch. This is easily undone if not needed.
- The check against abundance is now flipped so that higher abundance means a higher chance of catching an adult, rather than a juvenile. I assume the previous behaviour was a mistake, but if that was intended then it is also easily undone.
- The fishing pole's durability is now reduced after a successful catch, and when the rope is snapped, which fixes [#9241](https://github.com/anegostudios/VintageStory-Issues/issues/9241).
- Fish entities are now only caught if they like the current bait type on the bobber. Previously there was no check for bait type when catching an entity, so any fish that happened to go near the bobber would get caught, even if it wasn't actually lured to the bait.

#### Other refactoring
Most of these should fractionally improve performance, but they mainly affect readability and maintainability of the code itself.
- Replaced `EnumBobberState.NoCatch` with `EntityFishCatch`. `NoCatch` represents essentially the same state as `Baiting` and so was redundant - it was being set once in a block that would never be reached, and then never checked. Whereas `EntityFishCatch` is a valid state that wasn't previously included, which made it impossible to use switch statements to properly filter logic by state.
- Ticking logic in `onServerTick` and `TryCatchFish` is now fully switched by `bobberState` so everything is clearer and neater, instead of messy nested and overlapping if blocks.
- The code to give or spawn ItemStacks in `TryCatchFish` is now separated from the code determining what the drops should be, for way less repetition and more readable logic.
- In `getRandomFishEntityProperties`, the checks for valid climate, accepted bait type, and animal map density are now combined into one loop - this should be more efficient.
- The check for pond size is now done before calculating the abundance value. This means that tiny ponds quickly exit the method with an abundance level of 0. This in turn means that all exits of the method either have _both_ abundance of 0 and null return, _or_ non-zero abundance and non-null return. So either of those values can be used as a simple check for whether there's a catch available, and they don't need to be cross-checked or compared against each other.
- Removed the redundant method `hasCatchable`. It simply checked for either non-zero abundance or a caught entity fish. But its only call was when catching a non-entity fish, so it could be replaced with a bare abundance check.
- Removed a redundant check for `abundanceValue <= 0` immediately after it was set to a clamped positive value.
- Removed a redundant `accum` field that had no purpose.

### DIFF
--- a/Entities/EntityBobber.cs
+++ b/Entities/EntityBobber.cs
@@ -364,6 +364,9 @@ namespace Vintagestory.GameContent
                         World.SpawnItemEntity(drop, entityCatcher.Pos.XYZ);
                     }
                 }
+
+                ItemSlot slot = entityCatcher.ActiveHandItemSlot;
+                slot.Itemstack.Collectible.DamageItem(__instance.World, entityCatcher, slot);
             }
         }
 
--- a/Entities/EntityBobber.cs
+++ b/Entities/EntityBobber.cs
@@ -352,14 +352,17 @@ namespace Vintagestory.GameContent
                 }
             }
 
-            BaitStack = null;
-            WatchedAttributes.MarkPathDirty("baitStack");
-
-            foreach (ItemStack drop in drops)
+            if (drops.Length > 0) 
             {
-                if (!entityCatcher.TryGiveItemStack(drop))
+                BaitStack = null;
+                WatchedAttributes.MarkPathDirty("baitStack");
+                
+                foreach (ItemStack drop in drops)
                 {
-                    World.SpawnItemEntity(drop, entityCatcher.Pos.XYZ);
+                    if (!entityCatcher.TryGiveItemStack(drop))
+                    {
+                        World.SpawnItemEntity(drop, entityCatcher.Pos.XYZ);
+                    }
                 }
             }
         }
--- a/Entities/EntityBobber.cs
+++ b/Entities/EntityBobber.cs
@@ -433,7 +433,7 @@ namespace Vintagestory.GameContent
             abundanceValue *= pondSize / 1200f;
             if (printDebug)
             {
-                System.Diagnostics.Debug.WriteLine("Pond size: " + pondSize);
+                System.Diagnostics.Debug.WriteLine("3. Pond size: " + pondSize);
             }
 
             // 4. Now filter by fish depletion map
--- a/Entities/EntityBobber.cs
+++ b/Entities/EntityBobber.cs
@@ -109,7 +109,6 @@ namespace Vintagestory.GameContent
     public class EntityBobber : EntityProjectile, IRopeRippedListener
     {
         protected float swimmingAccum = 0f;
-        protected float accum = 0f;
         protected float catchAccum = 0f;
         protected float junkCatchChance = 0f;
         protected EnumBobberState bobberState;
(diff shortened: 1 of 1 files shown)


averixus:
Another issue identified and fixed: when catching entity fish, there was no check for the current bait type. This meant non-target fish could be caught by accident if they just happened to get near the bobber. Particularly frustrating if it was loaded with a high-value bait and a crappie swam past, but would also allow catching very large fish with poor or no bait. Now nearby fish are only caught if they would actually bite the current bait.

pizza2004:
Thanks for this!  These changes will be in a future version.

PR #196, opened 2026-06-06
Changed DialogueControllers' Dictionary for a ConcurrentDictionary in ConversableBehavior

Psyloh:
When, on a dedicated server, several players are talking to the same NPC, that's obvious a non-concurrent collection will fail (and it does)... That should fix it!

The error is just, when clicking answers, another player's controller can get retrieved and thus said controller issues a warning to flag the inconsistency...
That's easy to test at the treasure hunter's, as 2 players : both clicking quickly on answers until warnings proc (won't take long!)

### DIFF
--- a/Systems/Dialogue/BehaviorConversable.cs
+++ b/Systems/Dialogue/BehaviorConversable.cs
@@ -70,7 +70,7 @@ namespace Vintagestory.GameContent
 
             var controller = ControllerByPlayer.GetOrAdd(player.PlayerUID, _ =>
             {
-                var controller = new(world.Api, player, eagent, dialogueConfig);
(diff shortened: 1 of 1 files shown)


pizza2004:
Could you show an example of the warnings or link to an issue showing them?

Psyloh:
In SurvivalMod => DlgTalkComponent.SelectAnswerById:121
The Warning is logged erratically at server level cause the answer is null : client controller and server controller ain't on the same page ^^'
Only happens when 2 or more players are talking to the same npc : very easy to reproduce, just bring a friend to the treasure hunter and both quickly click on the answers randomly, you'll generate plenty ;-)
I sent a DM to you on Discord!

PR #199, opened 2026-06-18
Fix: mixed pies name derivation

### LINKED ISSUE #9461: Custom mixed pies show incorrect names on the first ingredient mix
### Game Version

v1.22.2

### Platform

Windows

### Modded

Modded

### SP/MP

Singleplayer

### Description

When a mod that adds a custom mixed pie is installed, making one shows an incorrect name on the first mix of the required ingredients.

In my test, mixing cabbages and fruits showed "Pot pie".
Since cabbage is a pot pie ingredient, that might be related.

### How to reproduce

[mixedpietest_v0.1.0.zip](https://github.com/user-attachments/files/27721882/mixedpietest_v0.1.0.zip)

Install this mod.
It adds a custom mixed pie recipe called "My mixed pie" (code: `mymixedpie`).
"My mixed pie" can be made by mixing cabbages and fruits (except for apples) in a pie.
A correct pie should look like:

Image

1. Enter a creative world
2. Take a table, 2 spelt doughs, 6 cabbages, and 2 redcurrants from the creative inventory
3. Place the table and put the doughs on it to start making a pie
4. Add 6 cabbages
5. Add 2 redcurrants

#### Expected behavior

A full "My mixed pie" is created.

#### Actual behavior

The appearance is correct, but the name is "Pot pie"

### Additional note

As I mentioned with "first mix", after the wrong name is shown, adding more ingredients (if possible) fixes the name.
For example, adding 4 cabbages and then 2 redcurrants shows "Pot pie", but adding 2 more redcurrants after that fixes the name, resulting in "My mixed pie".

### Screenshots

A custom mixed pie showing "Pot pie" as its name:

Image

### Logs (Required)

[Logs_.zip](https://github.com/user-attachments/files/27721986/Logs_.zip)

at375:
When a pie contains fillings from different food categories whose item types share mixing codes, the game determines a common mixed name by intersecting the mixing code arrays of all present fillings.

The original code performed the intersection in a loop using the mixing codes of the previous filling (still held in the stale variable) instead of the current filling being processed. Because of this, whenever a new filling would change the common set – for example, going from three fillings with mixing codes `[x, y]` to a fourth with only `[y]` – that final ingredient’s codes were effectively ignored. When adding filling 4 - intersection would be performed of filling 3 and intersected set of fillings 1 and 2, leaving the mix unchanged at `[x, y]`. The game then picked the first element (`x`) for the pie name, producing an incorrect result.

Fix corrects the loop so that intersection is performed on the mixCodes set with already added filling and mixing codes of the current filling, so the mixCodes set is immediately accurate.

Fixes [#9461](https://github.com/anegostudios/VintageStory-Issues/issues/9461)

Harmony patch mod version of the fix:
[Mod](https://mods.vintagestory.at/piefixes)
[Mod patch code](https://codeberg.org/at375/PieFixes/src/commit/e3464cf1cb5eba1d98af74409c0b0c04d86e1d2c/PieFixes/Patches/BlockPie.cs)

### DIFF
--- a/Block/BlockPie.cs
+++ b/Block/BlockPie.cs
@@ -275,8 +275,8 @@ namespace Vintagestory.GameContent
                 if (cStacks[i] == null) continue;
 
                 equal &= cstack.Equals(api.World, cStacks[i], GlobalConstants.IgnoredStackAttributes);
-                foodCatEquals &= cStacks[i] == null || foodCats[i] == foodCat;
-                mixCodes = cstack.ItemAttributes["inPieProperties"].AsObject<InPieProperties>()?.MixingCodes.Intersect(mixCodes) ?? [];
+                foodCatEquals &= foodCats[i] == foodCat;
+                mixCodes = cStacks[i].ItemAttributes["inPieProperties"].AsObject<InPieProperties>()?.MixingCodes.Intersect(mixCodes) ?? [];
 
                 cstack = cStacks[i];
                 foodCat = foodCats[i];


pizza2004:
Appreciate this!  We'll include this in a later version.

PR #200, opened 2026-06-23
Remove duplicate pickup WorldInteraction from pies

Lazula:
Removes the default meal RMB pickup WI in favor of the empty hand + RMB WI from BehaviorRightClickPickup.

### DIFF
--- a/Block/BlockPie.cs
+++ b/Block/BlockPie.cs
@@ -296,7 +296,7 @@ namespace Vintagestory.GameContent
 
             if (!foodCatEquals && mixCodes.Any())
             {
-                return Lang.Get("pie-mixed-" + mixCodes.First() + "-" +state);
+                return Lang.Get("pie-mixed-" + mixCodes.First() + "-" + state);
             }
 
             return Lang.Get("pie-mixed-" + FillingFoodCategory(cStacks[1]).ToString().ToLowerInvariant() + "-" + state);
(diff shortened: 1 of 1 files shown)


Block/BlockPie.cs
```diff
@@ -554,8 +554,8 @@ public override int GetRandomContentColor(ICoreClientAPI capi, ItemStack[] stack
 
         public override WorldInteraction[] GetPlacedBlockInteractionHelp(IWorldAccessor world, BlockSelection selection, IPlayer forPlayer)
         {
-            var baseinteractions = base.GetPlacedBlockInteractionHelp(world, selection, forPlayer);
-            baseinteractions = baseinteractions.RemoveAt(1);
+            WorldInteraction[] baseinteractions = [.. base.GetPlacedBlockInteractionHelp(world, selection, forPlayer)
+                .Where(bi => bi.ActionLangCode != "blockhelp-meal-eat" && bi.ActionLangCode != "blockhelp-meal-pickup")];
```
pizza2004, line 558: Is there even any point in grabbing this anymore?  Doesn't this remove the entire thing?  Or are you trying to get the behaviors from this?  I'm not sure this is the best approach.

Lazula: Well, there could theoretically be other interactions from BlockMeal, so I wanted to leave that open. It was previously just doing `baseinteractions.Remove(1)`, and this is at least more resilient than that. It might be better to just ignore all the interactions from BlockMeal and only use ones explicitly made for pie, but I'm not sure.

PR #201, opened 2026-06-23
Fix taking a slice from a pie block not costing tool durability

Lazula:
Using the crafting grid already does this.


pizza2004:
This will be in a future version, thanks!

PR #202, opened 2026-06-23
Fix dough held interaction not showing table stacks

Lazula:
Uses a `static WorldInteraction[]` instead of a `static ItemStack[]`

### DIFF
--- a/Item/ItemDough.cs
+++ b/Item/ItemDough.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
 using Vintagestory.API.Client;
 using Vintagestory.API.Common;
 using Vintagestory.API.Config;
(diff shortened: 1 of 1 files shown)


Item/ItemDough.cs
```diff
@@ -10,30 +11,28 @@ namespace Vintagestory.GameContent
...
 
-        public override void OnUnloaded(ICoreAPI api)
-        {
-            tableStacks = null;
-        }
+            ItemStack[] tableStacks = [.. api.World.Collectibles
+                .Where(obj => (obj as Block)?.Attributes?.IsTrue("pieFormingSurface") == true)
+                .Select(obj => new ItemStack(obj))];
 
+            interactions = [
```
pizza2004: Can this be ??=, since it's static so it shouldn't need to be done repeatedly?  In the same sense you can put the `tableStacks` variable directly into this part so it doesn't calculate it over and over for every dough item.

pizza2004:
Thank you!  Will be in a future version.

PR #203, opened 2026-06-23
Fix missing JsonItemStack.Code in BlockLiquidContainerBase.GetNutritionProperties

Lazula:
The code must exist when `FoodNutritionProperties.EatenStack` is cloned, or it crashes.


pizza2004:
This was fixed internally a different way already, but thank you!

PR #204, opened 2026-06-23
Tweak: Show total satiety in meal nutrition facts

Lazula:
Show the total satiety above the nutrition facts listing to prevent lots of mental math.


pizza2004:
Why does the apple pie give such little satiety?

Lazula:
That's the creative stack, which only has a single layer of filling. Here's a normal baked one. Sorry for the confusing screenshot!

PR #209, opened 2026-06-29
Normalize ignite world interaction generation, fix boiler nullable capi

KnewOne:
This PR fixes some world interaction generations trying to create the interactions in the server side, leading to null capi being passed around which didn't create issues since the part that used the capi was not activelly reachable for torch and firestarter, as those didn't have defined creative stacks. 
Adding ignitable items or changing existing items to have creative stacks akin to lantern would produce NPEs.
Additionally, i've taken the liberty of modifying the interaction generator code to use exclusively capi to avoid the repeat of this, as well as normalized the approach to branch execution: check against capi being null, or checks against api being castable to capi. Returns if the block is the last in the OnLoaded and IF blocks if it's indented and as such has worse visibility.
image

### DIFF
--- a/Block/BlockBloomery.cs
+++ b/Block/BlockBloomery.cs
@@ -16,16 +16,15 @@ namespace Vintagestory.GameContent
         {
             base.OnLoaded(api);
 
-            if (api.Side != EnumAppSide.Client) return;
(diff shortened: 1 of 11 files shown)


pizza2004:
Thanks!  Should be in a later version.

PR #210, opened 2026-07-03
Improved Spur Gears feature bounty.

Algernon733:
Adds placement logic to spur gears.

You can now place axles inside spur gears to allow mechanical power through the spur's axle input axis.
You can now place an extra spur gear perpendicularly in another spur gear by placing it on an adjacent axle.
 In total this allows you to send power from a spur gear to any of it's adjacent directions.

{6DCE526E-0CDC-4070-A53C-4A4606402D2C}

This comes with JSON changes in the assets folder. Since assets aren't in this repo you can download them here https://drive.google.com/file/d/1jKhaOI2VtG3A3kl8O5_Usy2N_KEb8Inm/view?usp=sharing
or I can send them another way.

### Followups
1: Should there be 3 way gears? Placing gears at all the arrows in your requirement image would result in a deadlock, but it does feel like this type of gear should allow 3 or 4 way configurations.
image

2: My current logic has one spur/axle item for each spur/axle that is shown in the block. The angled gear's logic has just 1 gear item for each block, and it updates the gear block based on the adjacent axles. In other words you need 2 spurs to create an angled connection with my spur logic, and 1 angled gear to create an angled connection with the angled gears logic.
Should we use my logic or the angled gear's logic? 
image

3. How should the different axle types work on the spurs? Currently only the wooden axle can be put on the spur.
image

4. For the extra spur, in addition to the current placement on the adjacent axles, should there be directional placement on the spur hitbox?
image

5. How should the connection to the spur look?
image

6. In order to mesh the spur-16 shape's gear teeth with adjacent spurs, I needed to move the spur halfway out of the block's boundary.
If that is the right approach, then how should the spur hitbox look?
image

### DIFF
--- a/Systems/MechanicalPower/Block/BlockAxle.cs
+++ b/Systems/MechanicalPower/Block/BlockAxle.cs
@@ -26,11 +26,14 @@ namespace Vintagestory.GameContent.Mechanics
             if (BlockSpurGear.WoodenAxleCheck(this) && world.BlockAccessor.GetBlock(blockSel.Position) is BlockSpurGear spurgear)
             {
                 if (spurgear.TryAddInsideAxle(world, blockSel.Position))
+                {
                     return true;
-
-                failureCode = "notreplaceable";
-
-                return false;
+                }
+                else
+                {
+                    failureCode = "notreplaceable";
+                    return false;
+                }
             }
 
             if (!CanPlaceBlock(world, byPlayer, blockSel, ref failureCode))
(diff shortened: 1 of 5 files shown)


Algernon733:
Turned off the ability to place spurs that create 4 way junctions because gears at the red positions in this video would deadlock

https://github.com/user-attachments/assets/dcff5f14-ed99-4cd0-8042-47522a5f8264

PR #212, opened 2026-07-03
Bug bounty - Improved Spur Gears

Zaldaryon:
## Multi-disc spur gear hub

Spur gears stack multiple discs on a single block. Each disc mounts on its own supported axle. The block redirects mechanical power between 2-6 axles without chaining angled gears.

Modifies `BEBehaviorMPSpurGear.cs` and `BlockSpurGear.cs` in place. Adds `BlockSpurGearMulti.cs`, `SpurGearMultiBlockRenderer.cs`, and two JSON blocktypes. Merge and it runs.

### How it works

1. Place a spur gear on an axle (vanilla placement, no change)
2. Place a second spur gear on the same block, clicking the face with another supported axle
3. Block morphs to `spurgearmulti` carrying both discs
4. Repeat for up to 6 faces
5. Break any axle: its disc drops, remaining discs stay
6. One disc left: block shrinks back to plain `spurgear`

Co-axial exits keep rotation direction. Perpendicular exits invert (same as angled gear). Each disc adds 0.0005 resistance.

Placement scans all 6 faces for a valid axle (clicked face gets priority, falls through to others). If the disc lands on a different face, the player gets a chat notification.

### Changes

| File | Edit |
|------|------|
| `BEBehaviorMPSpurGear.cs` | **Modified.** `discFaces` bitmask + TreeAttribute persistence, multi-disc `GetMechPowerExits`, resistance scaling |
| `BlockSpurGear.cs` | **Modified.** 6-face axle scan, redirect to `TryAddDisc` on existing gear, hub variant support |
| `BlockSpurGearMulti.cs` | **New.** Multi-disc block logic |
| `SpurGearMultiBlockRenderer.cs` | **New.** Instanced per-face disc rendering |
| `spurgearmulti.json` | **New.** Blocktype |
| `spurgearhub.json` | **New.** Hub variant (through-shaft) |

### Testing

PoC mod runs 31 assertions on headless server (placement, hub morph, multi-disc growth, disc drop, shrink-to-plain, resistance scaling, direction propagation).

[ImprovedSpurGears.zip](https://github.com/user-attachments/files/29632135/ImprovedSpurGears.zip)

Note: disconnected clutch discs still render spinning. Per-face angle freeze is possible but adds BE complexity. Left as-is.

### DIFF
--- a/Systems/Core.cs
+++ b/Systems/Core.cs
@@ -493,6 +493,7 @@ namespace Vintagestory.GameContent
             api.RegisterBlockClass("BlockAxle", typeof(BlockAxle));
             api.RegisterBlockClass("BlockAngledGears", typeof(BlockAngledGears));
             api.RegisterBlockClass("BlockSpurGear", typeof(BlockSpurGear));
+            api.RegisterBlockClass("BlockSpurGearMulti", typeof(BlockSpurGearMulti));
             api.RegisterBlockClass("BlockWindmillRotor", typeof(BlockWindmillRotor));
             api.RegisterBlockClass("BlockToggle", typeof(BlockToggle));
             api.RegisterBlockClass("BlockPulverizer", typeof(BlockPulverizer));
--- a/Systems/MechanicalPower/Block/BlockAxle.cs
+++ b/Systems/MechanicalPower/Block/BlockAxle.cs
@@ -22,8 +22,63 @@ namespace Vintagestory.GameContent.Mechanics
 
         public override bool TryPlaceBlock(IWorldAccessor world, IPlayer byPlayer, ItemStack itemstack, BlockSelection blockSel, ref string failureCode)
         {
+            BlockPos originalPos = blockSel.Position.Copy();
+            BlockFacing originalFace = blockSel.Face;
+            bool originalDidOffset = blockSel.DidOffset;
+            Block blockAtSelection = world.BlockAccessor.GetBlock(blockSel.Position);
+
             if (!CanPlaceBlock(world, byPlayer, blockSel, ref failureCode))
             {
+                if (failureCode != "notreplaceable") return false;
+
+                if (blockAtSelection is BlockSpurGear selectedGear)
+                {
+                    BlockFacing gearOrientation = BlockFacing.FromFirstLetter(selectedGear.Variant["orientation"]);
+                    BlockFacing hubFace = gearOrientation.Opposite;
+                    BlockPos hubAxlePos = originalPos.AddCopy(hubFace);
+
+                    Block blockAtHub = world.BlockAccessor.GetBlock(hubAxlePos);
+                    if (blockAtHub != null && blockAtHub.BlockId != 0 && !blockAtHub.IsReplacableBy(this))
+                    {
+                        failureCode = "notreplaceable";
+                        return false;
+                    }
+
+                    string axleRotation = (hubFace == BlockFacing.NORTH || hubFace == BlockFacing.SOUTH) ? "ns"
+                        : (hubFace == BlockFacing.UP || hubFace == BlockFacing.DOWN) ? "ud" : "we";
+                    Block axleVariant = world.GetBlock(new AssetLocation("game", "woodenaxle-" + axleRotation));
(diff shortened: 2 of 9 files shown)


Zaldaryon:
Pushed a second commit that reworks the rotation direction logic of the multi-disc gear. The first version handled placement and network membership, but the rotation senses were only correct for some entry directions: a gear driven from above spun its outputs correctly, the same gear driven from below did not, and results could change with the order the discs were placed in.

**Root cause**

I had modeled the perpendicular outputs like the large gear does, all of them turning in the same sense, opposite to the input. That works for the large gear because its rim gears sit on parallel axes. A multi-disc hub is different: perpendicular discs mesh like bevel gears at the edge shared by the two faces, so the output sense depends on which face the disc sits on, not just its axis. Two discs on opposite faces of the same axis have to counter-rotate each other. No per-axis convention can express that.

**What v2 does**

- Every exit sense derives from the entry path's actual rotation sense (`MechPowerPath.NetworkDir()`, which is stable across both the spread and the tryConnect calling conventions) via the invert flag. Co-axial discs continue the shaft unchanged.
- Perpendicular exits follow bevel kinematics: rim velocities match at the shared edge, so the sign flips with the geometric side of each face.
- `GetPropagatingTurnDir` and `IsPropagationDirection` answer with the same algebra, so a shaft connected to an already powered gear gets exactly the sense the initial network spread would have produced. This removes the placement-order dependence. The plain spur gear keeps the vanilla behavior.
- When the disc set changes on the propagation axis, the gear calls `MechanicalPowerMod.RebuildNetwork`, the same recovery the game runs on node removal. This is needed because `JoinAndSpreadNetworkToNeighbours` never revisits nodes already in the network, so already connected shafts would keep stale senses otherwise.

Two vanilla details that may be useful to know for other blocks: `BEBehaviorMPSpurGear.GetPropagatingTurnDir` returns `propagationDir.Opposite` unconditionally, which is the counter-rotation of a parallel meshed gear and the wrong sense (and wrong axis) for perpendicular exits, so the multi variant overrides it. And the network spread early-return mentioned above means any block whose exits change at runtime needs a rebuild to propagate that change.

One ambiguous case: a through-shaft (discs on both faces of one axis) plus a perpendicular disc is physically overconstrained. The through pair co-rotates as one shaft, but each end would demand the opposite sense from the perpendicular disc, so one mesh edge cannot look right no matter what. Real gearboxes avoid exactly this contact. I resolve it deterministically (the positive face of the axis counts as the entry), so the outcome never depends on build order.

**Validation**

I wrote a headless server-side test system that builds every combination: 6 entry faces x 31 disc subsets = 186 rigs, each fed by a creative rotor, asserting the exact propagation facing of every output shaft, plus the scope 1 tests from the first commit. All 217 pass. The senses were also verified visually in game against meshed rigs, including the case that first exposed the bug (a gear driven from below with discs on north, south and west). The test system only activates with the `SGTEST=1` environment variable set, so it is inert in normal play. I can share it or the run logs if useful.

Attached is a build of this as a standalone mod zip (with extra diagnostic logging left in) if you want to try the behavior in game without building the branch.

[ImprovedSpurGears-v2.zip](https://github.com/user-attachments/files/29651834/ImprovedSpurGears-v2.zip)

Zaldaryon:
On the physics question raised in Discord: yes, real spur gears only transfer torque between parallel shafts. Perpendicular transfer is bevel gear territory. The bounty spec image shows arrows going in multiple directions from a single block, which is physically a bevel/miter junction, not a spur mesh.

I treated the disc stacking as the gameplay abstraction it is: the player places spur gear items, the block handles the direction change internally. The rotation logic in v2 uses bevel kinematics under the hood (rim velocity matching at the shared edge between faces) so the visual spin directions are physically correct for that type of junction, even if calling it a "spur gear" is a stretch mechanically.

If you'd rather keep perpendicular exits out and limit the block to co-axial pass-through only (true spur behavior), that's a smaller subset of what's already here. I can strip the perpendicular logic into a separate PR or just delete it. Let me know which direction you want.

Zaldaryon:
Bug bounty is over.

PR #215, opened 2026-07-15
Don't display a "not filling" error when picking up a raw pie into a stack

Lazula:
Skip TryAddIngredientFrom if the pie is mergeable into the held item stack.


pizza2004:

PR #216, opened 2026-07-16
Fix crash when rendering pie with an invalid size

Lazula:
Prevents out-of-bounds indexing into `pieShapeBySize`.


pizza2004:

PR #217, opened 2026-07-16
Fix GenerateRandomPie not always using all food categories for mixed pies

Lazula:
This fix doesn't cover narrowing mixing codes down to keep it simple until the big changes in 1.23.


pizza2004:
We'll have this in a future version, thanks!

PR #225, opened 2026-09-18
Update Beehive OnScanForEmptySkep

danger-dan:
Now uses position aware block material and attributes, this allows 'dynamic' blocks (via entities) to act as 'beeFeed'.


danger-dan:
I don't know if this is all thats required - I can't test this. I might have missed something somewhere, but the idea stands

